PR #7, opened 2013-11-21
Added links to AutoFixture and other Ploeh repos,

ploeh:
although currently, Hyprlinkr has no open issues at all.


dahlbyk:
Thanks for these!

No open issues _may_ be a good problem to have...

PR #8, opened 2013-11-21
Added a link to original Jump in! article.

ploeh:
Since this page actually uses the "Jump in!" term, I thought it would be appropriate to link to the article introducing the concept.

If you don't like this suggestion, please feel free to reject the pull request.


dahlbyk:
I like it, at least as long as that text is there...

PR #12, opened 2013-11-21
Corrected Hyprlinkr ULR

ploeh:
since it seems I accidentally pasted the Albedo URL into that link as well as for Albedo :$

(Sorry)


dahlbyk:
I should have noticed that too

Issue #3, opened 2013-11-21
Write some words about getting involved as a contributor


shiftkey:
TODO: make with the words

dahlbyk:
Addressed by #14

PR #14, opened 2013-11-21
Elevator Pitch

### LINKED ISSUE #4: Write some words about getting involved as a maintainer
TODO: make good words

shiftkey:
Resolves #4 and #3 

I wanted it to be relatively simple, but also with some details around the process. Probably needs some massaging

Thoughts? @dahlbyk @jrusbatch @nikmd23 @avanderhoorn @haacked @shanselman

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,6 +46,21 @@
            <li>leave a message on a task you'd like to work on</li>
            <li>get to work!</li>
          </ul>
+
+        <h3>Projects</h3>
+
+        <ul>
+          <li><a href="https://github.com/Glimpse/Glimpse/issues?labels=Jump+In">Glimpse</a></li>
+          <li><a href="https://github.com/libgit2/libgit2sharp/issues?labels=Up+for+grabs">LibGit2Sharp</a></li>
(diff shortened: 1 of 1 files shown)


avanderhoorn:
Looks good to me, its simple and easy to understand. In 2 for contributors, I might be worth referencing that they could use tags (jump-in or up-for-grabs), milestones, etc.

shiftkey:
@avanderhoorn I added a mention about the tags here

DavidBurela:
I think the wording itself is alright. I would just modify the layout.
The 99% case is going to be people in the community coming here to see what they can contribute to, and how to do it.
The 1% case will be project maintainers coming here to find how they can add their project. They will read it once and then it isn't needed again.

I would format it as:
- How to get started
- List of projects

And then move the project owners details out to another page

shiftkey:
screen shot 2013-11-21 at 12 00 01 pm
screen shot 2013-11-21 at 12 00 18 pm

dahlbyk:
I approve of this message. :shipit: and we can adjust language from there?

shiftkey:
@dahlbyk it is done

dahlbyk:
:clap:

Issue #17, opened 2013-11-21
More project info


dahlbyk:
Before too many projects are added, should we move beyond a simple list? I'm thinking we can use a table with a few columns:
- Project (linked)
- up-for-grabs tag name (linked)
- Tags (e.g. Glimpse might use "web, diagnostics, asp.net, c#, javascript")
- Description/Notes

I figure we can eventually enable filtering of the ever-growing list based on tags.

Anything else?

shiftkey:
:thumbsup: to tags
:thumbsup: to brief description
:thumbsup: to filtering

akoeplinger:
:+1: sounds like a good idea

ChrisMissal:
Yes to all, but I would like to see it stay as pure HTML/CSS/JS. Can the "table" be a JSON file that is filtered/searched via JS?

dahlbyk:
Moving into JSON would work, but for now I think we could get away with an HTML table with an embedded microformat:

``` html
<table>
  <tr>
    <td><a class="proj" href="https://github.com/dahlbyk/up-for-grabs.net">up-for-grabs.net</a>
        <a class="label" href="https://github.com/dahlbyk/up-for-grabs.net/issues?labels=up-for-grabs">up-for-grabs</a></td>
    <td><p class="desc">A site to explore open source projects and jump in!</p>
        <p class="tags">javascript, oss, web</p></td>
  </tr>
  ...
</table>
```

ChrisMissal:
Oh yeah, certainly.

akoeplinger:
This sounds like a perfect fit: http://jekyllrb.com/docs/datafiles/

Will try how that works :smile: 

_edit_ looks like that only works for Jekyll sites, which this is not yet. Might however be a good idea to turn it into one.

dahlbyk:
@akoeplinger that seems like a good way to go, to more easily maintain consistent markup. I'll admit I've not actually used Jekyll - is that migration something you could take on?

dahlbyk:
I'm debating if it's worth separating out the project and label links - the project link makes it easier to view the readme, but maybe the description included here would be sufficient?

If we don't go with separate links, a list may be sufficient:

``` html
<ul>
    <li><a class="proj" href="https://github.com/dahlbyk/up-for-grabs.net/issues?labels=up-for-grabs">up-for-grabs.net</a>
        <p class="desc">A site to explore open source projects and jump in!</p>
[...]

PR #26, opened 2013-11-23
Added GitHubFlowVersion and GitReleaseNotes, also included tagline

JakeGinnivan:
I also included a tagline in preparation for #17
#17 is getting harder to do because whoever does it will have to go searching. Was wondering maybe a simple `link - tagline [tags]` format would do for now, then it can be converted easily, also makes it easier to experiment.

Can remove if it is an issue for now

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
 <html>
 
   <head>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
```
shiftkey, line 1: wat

JakeGinnivan, line 1: o_0

No idea..

dahlbyk:
> Was wondering maybe a simple link - tagline [tags] format would do for now

I think I'm settling on using classes like other microformats, so if you want to get a jump start on that you could do something like this:

``` html
<li><a class="proj" href="https://github.com/JakeGinnivan/GitHubFlowVersion/issues?labels=up-for-grabs&state=open">GitHubFlowVersion</a>
    - <span class="desc">Easy SemVer for projects using GitHub flow</span> [<span class="tags">github, semver</span>]</li>
```

JakeGinnivan:
Sounds good, will update

dahlbyk:
Note that the project list has been moved. You can just `push --force` over your existing branch.

PR #29, opened 2013-11-24
Added Jekyll, moved project data to seperate .yml file

akoeplinger:
This moves the data about projects into _data/projects.yml and renders it via Jekyll.

I tried to come up with some sane tags for the YAML markup, feel free to suggest better names.


_data/projects.yml
```diff
@@ -0,0 +1,144 @@
+- name: Glimpse
+  desc: Glimpse is a web debugging and diagnostics tool used to gain a better understanding of whats happening inside of your ASP.NET application.
+  site: https://github.com/Glimpse/Glimpse
+  tags: ASP.NET, Web, WebForms, Glimpse, diagnostics, performance, profiling, timing
```
akoeplinger: This could actually also be a list, e.g.

```
tags:
  - ASP.NET
  - Web
  - WebForms
  - Glimpse
  - ...
```

I guess that would make it easier in the future when the data is consumed by JS, but I'm not sure if this is really necessary right now...

shiftkey: Would doing that up-front help make data-entry and parsing easier in the future?

akoeplinger: I just tried using the "| jsonify" filter when rendering the YAML data. Using a list there renders it as a JSON array,  which looks like a lot easier to process in JS than the comma based version:
json

shiftkey: :heart:

shiftkey: Oh, and using the project name as a tag feels like duplication - shouldn't we just search on the tags AND the project name instead? 

cc @dahlbyk

akoeplinger: Yeah, the project name shouldn't be part of the tags, I just copied them from NuGet.org.

akoeplinger: Instead of a list, we can use an array (elements enclosed in `[ ]`), which is much more compact. PR updated.

dahlbyk:
Thoughts on styling issue/tags rather than labeling them? http://dahlbyk-demo.github.io/up-for-grabs.net/

shiftkey:
@dahlbyk 

Quick brain dump:
- emphasize the title a bit more (font a bit bigger, bold, that sort of thing)
- maybe wrap the "tag name" in quotes
- instead of boxes everywhere, just use the bottom border to separate items 
- did we settle on how we were going to group these projects - technology/OS/??? - can we mock that out in these visuals?
- there's probably some style issues but I Am Not A Good Designer

dahlbyk:
I'll hack around a bit more on the styling, though we could definitely benefit from the input of a professional.

> did we settle on how we were going to group these projects - technology/OS/??? - can we mock that out in these visuals?

I figure next step after we have a way to include tags is to use them. If we need to call out specific tags, we can, but it seems like a frequency-based presentation (cloud or sorted list or...) would bubble them to the top anyway.

dahlbyk:
Styles updated per feedback...what say you?

akoeplinger:
Looks good, though I have a feeling that newbies might not immediately notice the links to "up for grabs" items.
Maybe we can have a button similar to "View on GitHub" in the top right corner, to make it more noticeable?

dahlbyk:
Something like this?

dahlbyk:
I'm happy with this for now - opening another issue to fill in missing data.

akoeplinger:
Yeah, looks better. Thanks for merging!

dahlbyk:
Thank _you_ for doing the hard part!

PR #34, opened 2013-11-25
Add metadata for the Fixie project.

plioi:


dahlbyk:
May want to mention that it's a .NET/C# testing framework, at least in tags. Also, the upforgrabs name and link should match.

plioi:
@dahlbyk Fixed, thanks.

dahlbyk:
Danke

PR #48, opened 2013-11-25
Minor tweaks to NuGet Gallery entries...

analogrelay:
Based on some feedback from @jeffhand.

Thanks! This should be the last one, sorry!


dahlbyk:
Blah, merged the other before the feedback could be incorporated. Sorry...

analogrelay:
No worries @dahlbyk - It was all just icing and nit picking anyway :P. Thanks!

PR #49, opened 2013-11-26
Update index.html

kunaldhir:


dahlbyk:
Centering the footer does seem reasonable, but I'd rather not use presentational markup. Can you switch this to a CSS change, probably attached to `#footer_wrap`?

kunaldhir:
Fine. I'll do it using CSS.

shiftkey:
Closing this as #58 replaces it

PR #52, opened 2013-11-27
Added a link to NEST

Mpdreamz:


dahlbyk:
> should not leave github online editor open for days expecting things to stay the same

Indeed.

Issue #32, opened 2013-11-25
Filter by tag


dahlbyk:
We have tags now - let's support filtering by them!

I can't shake the thought that some sort of tag cloud would be neat, though that's certainly not a requirement. Or it could be added later. Along with dozens of other visualizations of the included projects! But I digress...

Random thoughts:
- As @akoeplinger mentioned [here](https://github.com/dahlbyk/up-for-grabs.net/pull/29#discussion_r7879552), Jekyll can render the projects as JSON so we don't have to parse the HTML.
- Filtering by a tag should probably count as a navigation event and be permalinkable.
- Should we support multiple tag filters? AND or OR?
- Filtering by special characters, e.g. `F#`, should work.

shiftkey:
> Filtering by a tag should probably count as a navigation event and be permalinkable.

+:100: to making it permalinkable

> Should we support multiple tag filters? AND or OR?

:thumbsup: - use OR because the sample of projects we have at the moment is so small (lol webscale)

> Filtering by special characters, e.g. F#, should work.

:thumbsup:

akoeplinger:
I guess this can be closed now?

dahlbyk:
> I guess this can be closed now?

Yes!

PR #56, opened 2013-12-03
Added filter by tags feature

chandu:
Implemented filter by tags with permalinkable urls (#32).
Libraries used: 
- [sammy.js](http://sammyjs.org/) (compatible with IE8+ and other browsers, hope that is fine.)
- [chosen](http://harvesthq.github.io/chosen/)
- [underscore.js](http://underscorejs.org/)

Added a jekyll generator `ProjectsDataGenerator` to generate json for the Projects list as a object with projects and tags properties (am new to ruby, not sure if the code is efficient enough). 
Currently rendering the projects as json inline with the HTML as the number of projects is not that high.

Sample Screenshot


shiftkey:
Just having a play with it locally, and my first reaction:

Yes!

Will have a closer review of the code, but I don't suspect I'll have any issues with it.

shiftkey:
> Currently rendering the projects as json inline with the HTML as the number of projects is not that high.

Seems reasonable. Any tips for transitioning to a better way once we hit that limit?

shiftkey:
@Chandu thanks for the contribution!

> (compatible with IE8+ and other browsers, hope that is fine.)

Totally fine with me.

I see some Jasmine references as well as the `SpecRunner.html` file - is this something we can run locally to ensure all the tests pass?

chandu:
Thanks for the merge Brendan. I just checked the live site and the projects
json is rendered as null and none of the projects are listed. Verified it
locally and its working. Can you please check if the deployment went fine?

shiftkey:
Right, I've found the problem:

http://jekyllrb.com/docs/plugins/

> GitHub Pages is powered by Jekyll, however all Pages sites are generated using the --safe option to disable custom plugins for security reasons. Unfortunately, this means your plugins won’t work if you’re deploying to GitHub Pages.
> 
> You can still use GitHub Pages to publish your site, but **you’ll need to convert the site locally and push the generated static files to your GitHub repository instead of the Jekyll source files**.

If you test it locally with the --safe flag you'll see the same behaviour:

```
jekyll serve --watch --safe
```

chandu:
Oh... no wonder it was working locally then. I was using jekyll server
--watch. Is building locally before deployment an option or shall I remove
the dependency on the plugin and move the tags map extraction  to client
side?

shiftkey:
> Is building locally before deployment an option or shall I remove the dependency on the plugin and move the tags map extraction to client side?

Whichever is less terrible.
- we could have a script, say `update-projects.rb` which updates `index.html` to include the new JSON.
- if you're up for rewriting the Jekyll plugin in JS so we don't need to "remember" to run the script and commit the changes...

chandu:
@shiftkey I will change the code to move the processing to JS and send a pull request later today.

dahlbyk:
Few things:
1. Thanks for your work on this!
2. You can preview your PR if you push to your fork's `gh-pages` branch and delete the CNAME file (just make that your last commit and don't include it in the PR). @shiftkey: Pages feature request: ignore CNAME on forks rather than failing the Pages build.
3. So the gist is that `jsonify` doesn't work on GH Pages? I thought it had been tested in https://github.com/dahlbyk/up-for-grabs.net/pull/29#discussion_r7879552 (or maybe that was just local, @akoeplinger?)

chandu:
@dahlbyk : 
Thanks for the information about the  gh-pages review tip.

I think `jsonify` works, it's the custom plugins added to `_plugins` folder that will not work in Github Pages. In this case I have added `projects_data_generator.rb` for extracting tags from the projects list.

akoeplinger:
@dahlbyk `jsonify` works on GH Pages, as it is one of the few plugins whitelisted by GitHub, other/custom plugins don't work.

@Chandu moving the processing to JS seems like a good idea.

chandu:
Submitted pull request #57 to address this issue.

PR #57, opened 2013-12-03
Removed ProjectsDataGenerator plugin and moved the tag extraction to js

chandu:
Removed the `projects_data_generator.rb` plugin and moved the tags extraction code to client side.
Fixed the projects list empty issue.


dahlbyk:
I've reverted #56 for now - can you please reset to the current `gh-pages` and recommit (`git commit -C 8e2a51d` to reuse existing commit metadata)?

At least while the list is smallish, thoughts on leaving the project list in HTML (SEO!) and then filtering it down with JS once a filter is selected?

I noticed this is live at http://chandu.github.io/up-for-grabs.net/ - thanks for that!

chandu:
@dahlbyk I have reset the gh-pages on my fork.
I am not clear with recommit part of your message. Can you please give some more details/steps (never used git commit with -C)?

When I run the command on my local repo with `gh-pages` as current branch below is the output I get:

`Chandu@local /g/Github/up-for-grabs.net (gh-pages)`
`$ git commit -C 8e2a51d`
`# On branch gh-pages`
`nothing to commit, working directory clean`
`

dahlbyk:
Now that your `gh-pages` is up-to-date with mine, you can do this:

``` bash
git checkout fix-remove-plugins-dependency
git reset --soft gh-pages
git commit -C 8e2a51d
```

This will:
1. Switch to your `fix-remove-plugins-dependency` branch
2. Point `HEAD` at `gh-pages`, but leave the index (and working directory) with all your changes
3. Commit the index (which still hasn't changed from when you switched to `fix-remove-plugins-dependency`) by reusing the commit information (message, author, timestamp) from 8e2a51d (the commit you pushed for this PR). Instead of hard-coding that SHA, you could also 1) use `fix-remove-plugins-dependency@{1}`, which uses the reflog to figure out where your branch was before the `reset`, or 2) use `ORIG_HEAD`, where Git will have stored the current `HEAD` before doing the `reset`.

On my machine, that series of operations looks something like this:

index.html
```diff
@@ -113,32 +99,44 @@ <h2 id="project_tagline">explore open source projects and <a href="http://nikcod
...
-    );
-    return arr;
-  }
-
-  $('.projects tbody').shuffle();
-})(jQuery);
+  var projects = {{ site.data.projects | jsonify }};
 </script>
+<script src="//cdnjs.cloudflare.com/ajax/libs/underscore.js/1.5.2/underscore-min.js"></script>
+<script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
```
dahlbyk: Maybe switch this to `//cdnjs.cloudflare.com/ajax/libs/jquery/1.10.2/jquery.min.js` to avoid an extra DNS lookup?

chandu:
@dahlbyk Thanks. I have applied the changes. You can verify the changes @ http://chandu.github.io/up-for-grabs.net

index.html
```diff
@@ -51,25 +53,9 @@ <h2 id="project_tagline">explore open source projects and <a href="http://nikcod
...
-                 {% endif %}
-                 {% if project.tags %}
-                 <p class="tags">{{ project.tags | join: ', '}}</p>
-                 {% endif %}
-             </td>
-           </tr>
-         {% endfor %}
-         </table>
-     
+         <div id="projects-panel">
```
dahlbyk, line 56: Does it Just Work™ if you make this a wrapper for the fully-populated `<table class="projects">...</table>`? I expect the table would immediately be replaced by a new shuffled, filtered table, much like it is now.

dahlbyk:
Could the tags be sorted by alpha (or bonus points for sort by frequency)?

index.html
chandu, line 56: I did not get your question ..

chandu:
Definitely .. Will add sorting based on the frequency and push the change.

chandu:
@dahlbyk Would showing the frequency next to the tag be a good idea?
Something like:
sort

index.html
chandu, line 56: Do you mean, let the table be rendered as usual (without client side templates) for initial load and when filters are applied render the table with js templates?

dahlbyk:
> @dahlbyk Would showing the frequency next to the tag be a good idea?

Yes!

index.html
dahlbyk, line 56: > Do you mean, let the table be rendered as usual (without client side templates) for initial load and when filters are applied render the table with js templates?

Yes, include the server-side table initially (so Google sees it), but then replace it with the shuffled list from JSON.

dahlbyk:
Other than the question about initial markup, I feel pretty good about this. @shiftkey?

chandu:
@dahlbyk Pushed the changes to the fix branch. Url to verify: http://chandu.github.io/up-for-grabs.net/#/

index.html
chandu, line 56: I think that's a good idea. If a permalink url with filters is used, the page would first display fully populated table which would be quickly replaced with the filtered version(wouldn't be noticeable to users.). We have to keep server side markup and client side markup in sync (liquid markup for server, and underscore template for client) though

shiftkey:
:shipit:

index.html
dahlbyk, line 56: > We have to keep server side markup and client side markup in sync (liquid markup for server, and underscore template for client) though.

Yeah, I'm not too worried about that, considering the project list is pretty much all there is to this. :)

Make it so!

chandu, line 56: Ok. Should I apply that change as a part of this pull request, or create a new one? I would prefer the latter, but will wait to hear back from you.

dahlbyk, line 56: Separate PR is fine

PR #59, opened 2013-12-05
Adding the lists to the <p> element 

kunaldhir:
I is done so that the list would look related to the upper description.

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,13 +66,13 @@ layout: default
 
          <p>
            What sort of tasks are a good fit?
-         </p>
 
          <ul>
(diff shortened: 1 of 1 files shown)


shiftkey:
Do you want to update this `<p>` tags here as well?

https://github.com/dahlbyk/up-for-grabs.net/blob/gh-pages/index.html#L46

dahlbyk:
> Do you want to update this <p> tags here as well?

:+1:

dahlbyk:
> Do you want to update this <p> tags here as well?
> 
> https://github.com/dahlbyk/up-for-grabs.net/blob/gh-pages/index.html#L46

Fixed this separately

PR #69, opened 2014-01-29
Adding SQLProvider

forki:
/cc @pezipink


shiftkey:
:boom: thanks @forki!

PR #71, opened 2014-02-03
Add Fredis

buybackoff:
Added Fredis project


dahlbyk:
I let go of my urge to alphabetize the list long ago, but I :clap: the effort.

buybackoff:
Thanks!

PR #80, opened 2014-03-21
Make each tag a link to itself

xdumaine:
Display each tag in the tags list under a project as a link to itself for easily filtering to a tag without scrolling up and using the search box.


dahlbyk:
Awesome!

PR #81, opened 2014-03-21
Improve small device layout

xdumaine:
Improve the layout on small devices by showing the table as block,
removing min width (prevent horizontal scroll), and add a font-size of
16 to inputs and selects. This prevents most phones from zooming in when
one of them is focused.


shiftkey:
Just pulled this down and tested it in an iPhone 3GS emulator.

happy-2

dahlbyk:
So responsive to the size of my device! That should be a thing.

:sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles:

shiftkey:
@dahlbyk this "internet" thing will never take off

PR #84, opened 2014-03-28
Added phpDocumentor as a participating project

mvriel:
I would like to have phpDocumentor added to the list of participating projects. With an active development team we intend to have a fair number of up-for-grabs issues and are constantly curating the list.


shiftkey:
Thanks!

PR #96, opened 2014-06-27
Update projects.yml

theburningmonk:
Added a couple of F# (and one Dart) projects that I really need help with maintaining and updating.


shiftkey:
Thanks!

PR #100, opened 2014-06-28
added Make# up-for-grabs

sapiens:


dahlbyk:
Not sure how these were overlooked, merging!

PR #105, opened 2014-07-24
Adding logary

haf:
Linking to logary

### DIFF
--- a/_data/projects.yml
+++ b/_data/projects.yml
@@ -1,3 +1,14 @@
+- name: logary
+  desc: |
+        Logary is a high performance, multi-target logging, metric and health-check library for mono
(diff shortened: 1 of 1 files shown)


_data/projects.yml
```diff
@@ -1,3 +1,14 @@
+- name: logary
+  desc: |
```
shiftkey, line 2: I'm rusty as hell on my YAML syntax - does this work, or can we just drop it?

haf, line 2: According to the spec it should work. I don't like lines longer than x chars. Your choice :)

shiftkey, line 2: :thumbsup:

shiftkey:
Thanks!

Issue #27, opened 2013-11-23
Figure out how to sort projects in the list


akoeplinger:
We need to talk about how we sort the projects (e.g. seems a bit unfair that Glimpse is always at the top of the list).

A few possibilites that came to my mind: 
- first-come, first served (as it is now)
- alphabetical
- popularity (might be tough to do)
- random

Thoughts?

JakeGinnivan:
Random would be my pick, or less popular were sorted at the top.

Sent from my Windows Phone

dahlbyk:
Once we have more metadata, I expect the full list to be hidden by default in favor of JS navigation by tag (cloud?). At that point we can do pretty much whatever we want - random feature, recently added (if we capture date), etc.

Looking up stars, issue count, etc. would also be handy, but is trickier without a server-side component.

itsjohncs:
When you click through to a project, then go back in your browser, the list is reshuffled. This makes it unpleasant to browse the list. Storing the seed in a cookie or local storage could solve this, or using a seed that only changes once a day, or encoding the seed in the url? I can code this up if there's some kind of consensus.

dahlbyk:
> When you click through to a project, then go back in your browser, the list is reshuffled. This makes it unpleasant to browse the list.

I'd never noticed that - good point!

> Storing the seed in a cookie or local storage could solve this, or using a seed that only changes once a day, or encoding the seed in the url? I can code this up if there's some kind of consensus.

Throwing a random seen in either `sessionStorage` or a cookie sounds like a good place to start.

itsjohncs:
Sounds good. Apparently you can't seed javascript rng so I just stored the ordering in a list.

dahlbyk:
Seems like this is sufficiently addressed?

PR #109, opened 2014-08-05
Prevents reshuffling within browser session.

itsjohncs:
With the current setup, the project list gets reshuffled if the user clicks through to a project then comes back. This makes it hard to browse the list. Now the ordering is stored in `sessionStorage` after the initial shuffling.
- If sessions storage is unavailable, the current behavior of shuffling every time is used.
- This patch assumes that project names are unique.
- See #27.
- I don't do a whole lot with javascript so am mostly unaware of best practices. This was also a hastily done patch, so please review!

Testing:
- Clicked through to a project and came back.
- Refreshed the page.
- Fooled browser into thinking sessionStorage was unavailable.

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -69,9 +69,32 @@
 
 	var ProjectsService = function (projectsData) {
 		var _projectsData = extractProjectsAndTags(projectsData);
-    var projects = _.toArray(_.shuffle(_projectsData.projects));
-		var tagsMap = {};
-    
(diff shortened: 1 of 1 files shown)


itsjohncs:
Actually, thought of a better way to do the shuffling. Sorry for the spam, let me take another wack at this.

dahlbyk:
For future reference, you can just `push --force` to the same branch and the PR will auto-update.

PR #110, opened 2014-08-05
Prevents reshuffling within browser session

itsjohncs:
With the current setup, the project list gets reshuffled if the user clicks through to a project then comes back. This makes it hard to browse the list. Now the ordering is stored in `sessionStorage` after the initial shuffling.
- If sessions storage is unavailable, the current behavior of shuffling every time is used.
- See #27.
- I don't do a whole lot with javascript so am mostly unaware of best practices. This was also a hastily done patch, so please review!

Testing:
- Clicked through to a project and came back.
- Refreshed the page.
- Fooled browser into thinking sessionStorage was unavailable.

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -69,9 +69,28 @@
 
 	var ProjectsService = function (projectsData) {
 		var _projectsData = extractProjectsAndTags(projectsData);
-    var projects = _.toArray(_.shuffle(_projectsData.projects));
-		var tagsMap = {};
-    
+    var tagsMap = {};
+
+    var canStoreOrdering = (JSON && sessionStorage && sessionStorage.getItem
+                            && sessionStorage.setItem);
+    var ordering = null;
+    if (canStoreOrdering) {
+      ordering = sessionStorage.getItem("projectOrder");
+      if (ordering) {
+        ordering = JSON.parse(ordering);
+      }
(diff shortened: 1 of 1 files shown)


itsjohncs:
This differs from the last patch in that I changed from storing the project names to storing a list of indices from the static list. This cleaned up things considerably and removed the assumption that project names are unique.

akoeplinger:
What happens if a new project is added to the list? If I understand the code correctly, the new item wouldn't show up as its index isn't stored?

itsjohncs:
Correct, I figured that would be fine since the project would show up once `sessionStorage` gets cleared. What do you think?

itsjohncs:
Ooo, though if a project is taken out it'll error. That's a bigger problem, I'll just check that the sizes of the two arrays match and if they don't, reshuffle.

akoeplinger:
Ah yes, this uses `sessionStorage` so it is cleared when the browser/tab is closed (I've just been working with `localStorage` which is persistent, so I was confused here). I think this is fine then :smile:

itsjohncs:
Ping

shiftkey:
Seems legit. 

I'll give @dahlbyk a chance to look over it as well before hitting the button.

dahlbyk:
Of this I approve. Thanks!

PR #119, opened 2014-09-06
Added OpenRA to the list

Mailaender:
We aim to be very beginner-friendly. Join our http://www.openra.net/community/


shiftkey:
Welcome aboard!

PR #120, opened 2014-09-07
Added a missing space

Mailaender:
Something did not work. I assume a syntax/indention error.

### DIFF
--- a/_data/projects.yml
+++ b/_data/projects.yml
@@ -605,9 +605,9 @@
     link: https://github.com/adobe/brackets/labels/starter%20bug
 
 - name: OpenRA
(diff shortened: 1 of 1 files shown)


Mailaender:
Followup of https://github.com/dahlbyk/up-for-grabs.net/pull/119

shiftkey:
Weird, I can't see it in the list but I also didn't think the lack of a space would be a big issue.

Let's see what happens next...

Mailaender:
Still not there. I am pretty clueless what I did wrong.

akoeplinger:
The issue is the `:` in `Conquer: Red`, you'll need to enclose the description in quotes or YAML thinks it's a map.

PR #146, opened 2015-01-17
Added .NET Core to the project list

shiftkey:
They're using the tag, so why not add it here? :trollface:

@davkean @terrajobst let me know if you have a better suggestion for the description field.

### DIFF
--- a/_data/projects.yml
+++ b/_data/projects.yml
@@ -724,3 +724,11 @@
   upforgrabs:
     name: help out!
     link: https://github.com/kellabyte/Haywire/labels/help%20out!
+    
+- name: .NET Core
+  desc: The class libraries for the next version of the .NET Framework
+  site: https://github.com/dotnet/corefx/
+  tags: [ .NET ] 
+  upforgrabs:
+    name: up-for-grabs
+    link: https://github.com/dotnet/corefx/labels/up-for-grabs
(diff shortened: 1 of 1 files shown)


shiftkey:
There's an `accepting-prs` label which [appears to represent the same thing](https://github.com/dotnet/corefx/labels/accepting%20prs) - should this be replaced?

terrajobst:
Thanks! I meant to create the PR myself but lost track of it.

`Accepting PRs` is marking [API proposals](https://github.com/dotnet/corefx/wiki/API-Review-Process) that passed our review. Could be considered the same, though. [I've started a thread](https://github.com/dotnet/corefx/issues/455).

_data/projects.yml
```diff
@@ -724,3 +724,11 @@
   upforgrabs:
     name: help out!
     link: https://github.com/kellabyte/Haywire/labels/help%20out!
+    
+- name: .NET Core
+  desc: The class libraries for the next version of the .NET Framework
```
terrajobst: That's a bit misleading as .NET Core is a self-contained stack. Some of the API will be ported to the .NET Framework, some will not. What about: `Contains the foundational libraries that make up the .NET Core development stack`

shiftkey: :ok_hand: I wrote it in haste, not attached to it at all

terrajobst: Sweet! Thanks a lot for doing this!

PR #149, opened 2015-01-24
Update projects.yml

mcraken:
I added a project of mine to the list. Thanks in advance.

### DIFF
--- a/_data/projects.yml
+++ b/_data/projects.yml
@@ -1,3 +1,11 @@
+- name: Platform
+  desc: A platform to create a scale-able and container-less application. The platform is built on OSGi, Vertx.io, Cassandra, Ehcache, SOLR, Spring data, LMAX disruptor and Quartz.
+  site: https://github.com/mcraken/platform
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks @mcraken!

It'd be great if there were a list of issues on the repository of features/bugs/suggestions that could be contributed to the project by others. Would that be possible to do, so visitors don't land on the page and find themselves uncertain of what they can help out with?

mcraken:
Thanks for you comment @shiftkey 

Sure I would do what you have asked, but I'm already building a readme for the repository. I'm trying to explain the project and I'm providing some sort of a to-do list at the end.

Would you kindly check it out and suggest what to do or how to optimize it?

shiftkey:
@mcraken we added some notes to the end of the home page:

---

### What sort of tasks are a good fit?
- Tasks should take no longer than a few nights' worth of work
- Tasks should stand alone - avoid core functionality on which other tasks depend
- Tasks should be well described with pointers to help the implementer

mcraken:
@shiftkey Ok, working on it.

shiftkey:
We've made some changes to how projects are added in #160 - once you've got these tasks fleshed out I'd love to add the project to the list!

PR #168, opened 2015-02-20
Added BamPi

yannisgu:


yannisgu:
Oh, you are working on the new structure, I will resend it there...

PR #185, opened 2015-04-13
Add note to readme about creating file in browser

akoeplinger:
This should make it easier for folks who don't know that GitHub's UI also supports creating files.
Follow-up to https://github.com/dahlbyk/up-for-grabs.net/pull/160#issuecomment-92147715.

/cc @shiftkey


shiftkey:
@akoeplinger fantastic. thanks!

PR #187, opened 2015-04-14
Remove the extra "---" from top of the project files

### LINKED ISSUE #186: Splitted data files contain --- at top
E.g. [angularjs.yml#L1](https://github.com/dahlbyk/up-for-grabs.net/blob/d70d4b6ffcd14bc89a8c88a9ff76456a47b11b1d/_data/projects/angularjs.yml#L1)

This looks like a mistake from the migration, we should just remove the `---` (unless it's needed for something?).

akoeplinger:
This is a left-over from the migration and just ugly.
Fixes #186


shiftkey:
:sparkles:

PR #189, opened 2015-04-16
Add Backdrop CMS to projects available on up-for-grabs.net.

jenlampton:
Thank you so much for this great project.  We would love to be a part of it!


shiftkey:
Thanks for adding your project!

PR #194, opened 2015-04-29
Rename .NET project files

dahlbyk:
Lost the .NET projects, presumably because of the leading `.` in the filename. Giving this a try...


shiftkey:
Thanks for spotting this!

PR #200, opened 2015-06-30
Add fxor tool

abderraouf-adjal:


_data/projects/fxor.yml
```diff
@@ -0,0 +1,12 @@
...
+site: https://abderraouf-adjal.github.io/fxor
+tags:
+- C
+- Crypto
+- Files
+- Utils
+- Command-line
+upforgrabs:
+  name: fxor
+  link: https://github.com/abderraouf-adjal/fxor
```
shiftkey, line 12: This link should be to a list of issues which are suitable for new contributors, e.g:

https://github.com/abderraouf-adjal/fxor/labels/help%20wanted

_data/projects/fxor.yml
```diff
@@ -0,0 +1,12 @@
...
+desc: A tool to encrypt/decrypt a file using XOR operation to do one-time pad.
+site: https://abderraouf-adjal.github.io/fxor
+tags:
+- C
+- Crypto
+- Files
+- Utils
+- Command-line
+upforgrabs:
+  name: fxor
```
shiftkey, line 11: Instead of using the project name here, this should be the name of the label you use for tagging issues.

We suggest `up for grabs` but you can see from the project list that not everyone uses it...

PR #208, opened 2015-07-26
Prevent XSS

skazantsev:
I've noticed unescaped values in an underscore template which results in XSS if a project file contains malicious code e.g. &lt;img src="images/blacktocat.png" onload="alert('Hello XSS')">.

Before fix:
exploiting

After fix:
after-fix


shiftkey:
@skazantsev thanks for calling this out!

PR #209, opened 2015-07-27
Create nativescript.yml

valentinstoychev:


_data/projects/nativescript.yml
```diff
@@ -0,0 +1,10 @@
+name: NativeScript
+desc: Cross platform Native mobile apps with JavaScript
+site: https://nativescript.org
+tags:
+- JavaScript
+- mobile
+- native
+upforgrabs:
+  name: PRs plz!
+  link: https://github.com/nativescript/nativescript/labels/up-for-grabs
```
shiftkey, line 10: This URL should point to an existing tag on the project - unfortunately `up-for-grabs` doesn't exist on their repo.

Perhaps you could ask the team if this is something they're interested in...

_data/projects/nativescript.yml
```diff
@@ -0,0 +1,10 @@
+name: NativeScript
+desc: Cross platform Native mobile apps with JavaScript
+site: https://nativescript.org
+tags:
+- JavaScript
+- mobile
+- native
+upforgrabs:
+  name: PRs plz!
```
shiftkey: generally we name this the same as the label on the repo - this was separated out so that we didn't need to do URL parsing on the client.

_data/projects/nativescript.yml
valentinstoychev, line 10: This is done now - https://github.com/nativescript/nativescript/labels/up-for-grabs

Thanks for helping!

_data/projects/nativescript.yml
shiftkey: Could we rename this to be `up-for-grabs`?

_data/projects/nativescript.yml
shiftkey, line 10: :thumbsup:

_data/projects/nativescript.yml
valentinstoychev: Yes, I updated it. Is it OK now?

shiftkey:
Thanks!

Issue #76, opened 2014-02-07
Quick Links for Languages/Popular Tags


shiftkey:
It should be easy to quickly navigate projects based on common tags.

cc @jrusbatch

ChrisMissal:
How do you see this working @shiftkey? Links to the top __ tags? Links that pre-populate the filter?

I was going to take a stab at this, but found myself guessing too much.

shiftkey:
@ChrisMissal I'm not sure - it was so long ago when I logged this, and past me was vague and unhelpful.

What if we had a short list of the most common tags below the filter box - here's a crude mockup:

<img width="746" alt="screen shot 2015-07-28 at 11 36 43 am" src="https://cloud.githubusercontent.com/assets/359239/8922109/d0ac13fe-351d-11e5-8201-bfb5c7857693.png">

When you click on the tag, the filter gets pre-populated and then the filtering should kick in.

ChrisMissal:
Excellent, this was my initial thought as well.
![selfie-0](http://i.imgur.com/y4O1sYh.gif)

mikecole:
@ChrisMissal that beard.

michaelybecker:
Hi! Looking for up-for-grabs's but this seems wonderfully implemented. Is there a reason this issue isn't closed? (Pardon my n00bness, general disclaimers apply etc.)

shiftkey:
@MichaelHazani nope, no good reason at all! Thanks for reminding me!

PR #213, opened 2015-07-28
add popular tags section

ChrisMissal:
## Screenshot

image
## Usage

When links are clicked, it resets the filter to that tag. I chose 10 as a general fallback on the function, but for this layout, we'll want to show the top 4 most popular tags.

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -107,6 +107,10 @@
     this.getTags = function() {
       return tagsMap;
     };
+
+    this.getPopularTags = function (popularTagCount) {
+      return _.take(_.map(tagsMap, function (val) { return val; }), popularTagCount || 10);
+    }
   };
 
   host.ProjectsService = ProjectsService;
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -25,7 +25,7 @@
   var TagBuilder = function(){
     var _tagsMap = {},
         _orderedTagsMap = null;
-    
+
     this.addTag = function(tag, projectName){
       var tagLowerCase = tag.toLowerCase();
       if(!_.has(_tagsMap, tagLowerCase)) {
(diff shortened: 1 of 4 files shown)


ChrisMissal:
Closes #76

ChrisMissal:
You can preview this here: http://chrismissal.com/up-for-grabs.net

shiftkey:
In hindsight, I'm not sure how I feel about the popular tags being a larger font size than the project name. 

What if we made it a bit smaller? That means we could show a few more tags.

ChrisMissal:
image

ChrisMissal:
This size gives us 6 tags @shiftkey

ChrisMissal:
@shiftkey @dahlbyk @jrusbatch what do you guys think? (preview: http://chrismissal.com/up-for-grabs.net/#/)

javascripts/projectsService.js
```diff
@@ -107,6 +107,10 @@
     this.getTags = function() {
       return tagsMap;
     };
+
+    this.getPopularTags = function (popularTagCount) {
+      return _.take(_.map(tagsMap, function (val) { return val; }), popularTagCount || 10);
```
dahlbyk: Use `_.values(tagsMap)`?

dahlbyk:
Styling is a little weird on small viewports:
image

dahlbyk:
Thoughts on rendering the count in parens, like we do in the dropdown list?

ChrisMissal:
Good suggestions @dahlbyk, I updated the PR.

image

dahlbyk:
This works for me - any pre-closing thoughts?

jrusbatch:
@ChrisMissal Would you mind rebasing this?

ChrisMissal:
@jrusbatch not at all! but clearly i just did a bad thing and those commits are lost... I'm looking for them now

ChrisMissal:
Found em! Here you go @jrusbatch
/cc @dahlbyk @shiftkey

shiftkey:
Just had another look at this over at http://chrismissal.com/up-for-grabs.net/#/ - did we lose the parens?

ChrisMissal:
We sure did, likely I goofed during the rebase. They're back now.

shiftkey:
@ChrisMissal cool, glad to know I'm not going crazy (or crazier?)

PR #217, opened 2015-08-06
Clarify tags in README.md

The-Compiler:
When I first read it I didn't understand what the difference between `upforgrabs -> name` and `tags` is.


shiftkey:
@The-Compiler looks good to me!

PR #218, opened 2015-08-06
Add qutebrowser

The-Compiler:
Are those too many tags? I added tags with major versions because for all of those they're incompatible, and often referred to with the name and number.


hackebrot:
Cool :smile:

shiftkey:
> Are those too many tags?

You probably don't need to include all the versioned tags (perhaps `python3` is fine though).

The-Compiler:
I removed all of them (the versioned ones, that is) - I made the mistake to not look at existing tags before deciding what I should add :wink:

PR #222, opened 2015-08-09
adding Snowdrift.coop to up-for-grabs list

wolftune:
Using our homepage and aggregated ticket link although we also are on GitHub. We don't want to emphasize GitHub, but if there were a _separate_ way to link to both the homepage and to GitHub, we might do that.

### DIFF
--- a/_data/projects/snowdrift
+++ b/_data/projects/snowdrift
@@ -1,0 +1,15 @@
+name: Snowdrift.coop
+desc: Sustainable funding platform for free/libre/open works
+site: https://snowdrift.coop/
+tags:
+- oss
+- floss
+- web
+- yesod
+- haskell
+- javascript
+- html
+- css
+upforgrabs:
+  name: newbie-friendly
+  link: https://snowdrift.coop/p/snowdrift/t?_hasdata=&f1=newbie-friendly


shiftkey:
> Using our homepage and aggregated ticket link although we also are on GitHub. 

Totally fine :thumbsup:

> here were a _separate_ way to link to both the homepage and to GitHub, we might do that.

Could you elaborate on what you'd like to see supported here? I'm not sure I understand...

wolftune:
"if there were" not "here were" — you cut off part of the quote. Anyway, I just meant that if there were one entry for "homepage" and another for "repository" I might add links to both, but the homepage itself links to the repo, so that's fine. I just wasn't sure how GitHub-centric this was. Happy for it _not_ to be as some people connected with our project object to GitHub since it is proprietary itself…

shiftkey:
It's not intended to be GitHub-centric - URLs which do the equivalent features to a repository's issue list or a home page are totally fine here.

I guess what's unclear here is how to navigate from the project page to then start contributing. For the most part, I think projects leave that up to the user - I'm not sure what value adding in a `repository` link here would add, but I'm open to suggestions...

shiftkey:
@wolftune I'm happy with everything else in this PR.

If you'd like to talk further about the "repository URL" discussion please open an issue.

wolftune:
Thanks for the reply. I get it overall. I think for _us_ anyway, more useful than a dedicated repo link would be a link to our "how-to-help" page… but truthfully, I think it's a fine start. No need to discuss further here.

Incidentally, just for reference, you're overlapping largely with the well-known site http://openhatch.org/ and I'm skeptical of needless fragmentation — but that said, your site let us just link to our appropriate page, while OpenHatch only works with a limited set of specified supported APIs and we need to do extra work to add support for them, so nice that your site is more flexible in that respect.

PR #245, opened 2015-10-07
Add LeafletJS

IvanSanchez:


_data/projects/leaflet.yml
```diff
@@ -0,0 +1,10 @@
+name: Leaflet
+desc: JavaScript library for mobile-friendly interactive maps
+site: http://leafletjs.com
+tags:
+- Javascript
+- Maps
+- GIS
+upforgrabs:
+  name: easy-fix help-wanted
```
shiftkey: When we're parsing this content to generate the site, we expect to have just one value here.

I'm not sure which one you're more fond of, but if you could cut it back to one here it'd be :heart:

IvanSanchez: Well, there might be an `easy fix` bug which one of the hardcore developers wants to work with, and there might be a `help wanted` bug which just tries to find a contributor with the right browser configuration or hardware.

If I'd have to choose between the two of them, I'd probably say "keep easy fix".

Shall I amend this PR?

shiftkey: Yeah, it's been a long time since the site content had a good refresh - I need to sit down and rework the initial vision we had for this project. Our target audience is mainly people who want to get started contributing to a project, but the "needs specific environment" isn't something I've actively thought about.

> Shall I amend this PR?

Yes please!

_data/projects/leaflet.yml
```diff
@@ -0,0 +1,10 @@
...
+  link: https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+label%3A%22easy+fix%22+label%3A%22help+wanted%22
```
shiftkey: You could also tidy up your URL here to point to just the single tag

IvanSanchez:
OK, amended.

_data/projects/leaflet.yml
```diff
@@ -0,0 +1,10 @@
+name: Leaflet
+desc: JavaScript library for mobile-friendly interactive maps
+site: http://leafletjs.com
+tags:
+- Javascript
+- Maps
+- GIS
+upforgrabs:
+  name: easy-fix
+  link: https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+label%3A%22easy+fix%22
```
shiftkey, line 10: A bit of :lipstick: - this can be swapped out for https://github.com/Leaflet/Leaflet/labels/easy%20fix

PR #254, opened 2015-10-17
Create SharpBucket.yml

MitjaBezensek:


_data/projects/SharpBucket.yml
```diff
@@ -0,0 +1,13 @@
...
+site: https://github.com/MitjaBezensek/SharpBucket
+tags:
+- ".NET"
+- "C#"
+- Bitbucket
+- wrapper
+- rest
+- api
+upforgrabs:
+  name: easy fix
```
shiftkey: This should be `easy-fix` to match up with the tag name on the repo

PR #255, opened 2015-10-21
ElasticElmah

wallymathieu:
ElasticElmah is focused on ease of use. Mainly used for error logging (that's why synchronous requests rather than async or udp). It was made before kibana et.c. became popular (so it should be changed to log in that format instead).

### DIFF
--- a/_data/projects/ElasticElmah
+++ b/_data/projects/ElasticElmah
@@ -5,7 +5,7 @@ tags:
 - ".NET"
 - C#
 - Ruby
-- log
+- logging
 upforgrabs:
   name: up_for_grabs
   link: https://github.com/wallymathieu/ElasticElmah/labels/up_for_grabs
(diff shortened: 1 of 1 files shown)


_data/projects/ElasticElmah
```diff
@@ -0,0 +1,11 @@
+name: ElasticElmah
+desc: Logging log4net objects to elastic search from .net and log4r objects from ruby. The focus is on error logging, thus collecting as much information as possible when something is logged.
```
shiftkey: `elastic search` -> `Elasticsearch` ?

wallymathieu: Yes, spelled wrong.

_data/projects/ElasticElmah
```diff
@@ -0,0 +1,11 @@
...
+site: https://github.com/wallymathieu/ElasticElmah
+tags:
+- ".NET"
+- C#
+- Ruby
+- log
```
shiftkey: Some other projects use `logging` for a label. Might be easier for searching...

shiftkey:
Thanks!

PR #257, opened 2015-10-25
Add Serilog - Simple .NET logging with fully-structured events

augustoproiete:
Simple .NET logging with fully-structured events. Serilog combines the best features of traditional and structured diagnostic logging in an easy-to-use package

http://serilog.net


shiftkey:
Thanks!

PR #266, opened 2015-11-02
more automation, less mistakes

shiftkey:
- [x] wireup Travis to run against PRs
- [x] see build errors because of files not being marked with `.yml`
- [x] validate a bunch of attributes
- [x] actually exit if an error is found
- [x] catch duplicate tags
- [x] catch malformed tags


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: ruby
+rvm:
+  - 2.2
+
+sudo: false
+
+install:
+  - bundle install --path vendor/bundle
```
akoeplinger: add `bundle config build.nokogiri --use-system-libraries` before that one to avoid building nokogiri during build.

shiftkey: :metal: done further down in the file

shiftkey:
Awful ruby code aside, that was easier than I thought it'd be...

shiftkey:
Ready for review, but given this is not part of the core I'm not sure how thorough it needs to be...

PR #260, opened 2015-10-29
Hipster-style

daveaglick:
So I was sitting around listening to Mumford & Sons and contemplating my beard, when this happened.

In all seriousness I was mainly just playing around. I realize it's a big departure from the current style, so take it or leave it as you see fit. That said, if you think there's promise here I can tidy it up and add a little extra umph (some work needs to be done on responsive support, cleaning up unused CSS, styling headers, etc.)

Here's what it looks like right now:

2015-10-28_22h41_55

PR #267, opened 2015-11-02
Move the information for project mantainers at the top

luciomartinez:
Currently, there is useful information being show at the bottom of the site. That part is hard to reach, you have to scroll through the list of _several_ projects to see it. Moving it at the top, since it's not large, gives it better visibility.


shiftkey:
Thanks for submitting this. Unfortunately I'm not going to merge this, so I figure I should explain a bit about what I'm thinking here.

I Have Plans™ about reworking the words on the home page which will make this somewhat redundant. I'd like the header to contain a quick introduction for the site, and then link to other resources for people to read - rather than bring all the words up above the project list. 

So I think that combined with the work occurring in #260 will give us a better long-term solution to this.

PR #270, opened 2015-11-24
Added GlPortal.

hhirsch:


_data/projects/GlPortal.yml
```diff
@@ -0,0 +1,13 @@
...
+tags:
+- c++
+- python
+- javascript
+- Game
+- engine
+- 3d
+upforgrabs:
+  name: up-for-grabs
+  link: https://bugs.glportal.de/
```
shiftkey: It'd be great if this URL pointed to a filtered list of issues which new contributors might be able to get started on. Is this possible?

hhirsch:
I have read the instructions. We have issues tagged up-for-grabs, but our bugtracker does not support filtering for them. I am willing to create a seperate list if it is necessary. Let me know what you think.

shiftkey:
@hhirsch that URL is where the users will land after clicking on the project link, so if it's possible to make it less intimidating then I'd love to see that happen...

hhirsch:
@shiftkey So flyspray can't filter by tag. But I found another way to filter for issues. I have created a new task type called "Simple Task". Now people will find a simple small list.

hhirsch:
@shiftkey :white_check_mark: **It should meet your specs now.** You'll find specially selected issues under the url now. I will also add a bunch more so that there are always about 7 tasks from very easy to easy.

_data/projects/GlPortal.yml
```diff
@@ -0,0 +1,13 @@
...
+tags:
+- c++
+- python
+- javascript
+- Game
+- engine
+- 3d
+upforgrabs:
+  name: Simple Task
+  link: https://bugs.glportal.de/index.php?string=&project=1&type%5B%5D=5
```
shiftkey: You can drop the `string=&` parameter from here for a bit of extra :gem:

shiftkey:
@hhirsch so good! :sparkles: just one little suggestion!

hhirsch:
@shiftkey Hello. Shortened the URL.

shiftkey:
Thanks!

PR #272, opened 2015-11-25
Add Mozilla Investigator project

jvehent:
MIG is a platform to perform investigative surgery on remote endpoints. It enables investigators to obtain information from large numbers of systems in parallel, thus accelerating investigation of incidents and day-to-day operations security.
http://mig.mozilla.org


shiftkey:
Thanks!

PR #275, opened 2015-11-30
added Rebus project

mookid8000:


_data/projects/rebus.yml
```diff
@@ -0,0 +1,24 @@
...
+- "Azure Service Bus"
+- "Azure Storage Queues"
+- "distributed"
+- "messaging"
+- "service bus"
+- "soa"
+upforgrabs:
+  name: up-for-grabs
+  link: https://github.com/rebus-org/Rebus/labels/up-for-grabs
+stackoverflow:
```
shiftkey: cc @dahlbyk @jrusbatch for feels on extra topics (I like the SO option)

jrusbatch: I like SO, too. I'm not sure about `blog` though, that seems like it could overlap with `site`.

shiftkey: That seems fair. What about, as a first step, we make `stackoverflow` an optional element (I can spec it out as part of this review) and then work out how to surface it through the UI?

_data/projects/rebus.yml
```diff
@@ -0,0 +1,24 @@
...
+  name: Rebus questions on StackOverflow
```
shiftkey: @mookid8000 I don't think we need this element (unless we do something more fancy than link to SO). Thoughts?

_data/projects/rebus.yml
```diff
@@ -0,0 +1,24 @@
...
+  link: http://stackoverflow.com/questions/tagged/rebus
+blog:
```
shiftkey: As mentioned above, I don't know if this adds much value on top of the `site` link above.

dahlbyk: Agreed

_data/projects/rebus.yml
dahlbyk: I don't think we need it

jrusbatch: Agreed.

mookid8000:
I have removed the extra links now :smiley:

dahlbyk:
:sparkles:

mookid8000:
:zap: :clap: :tada:

PR #276, opened 2015-11-30
Correct site of Mutability Detector

Grundlefleck:
Without the `http://` prefix it looks like the value for site is generated relative to up-for-grabs.net, and of course, not resolving.


shiftkey:
Thanks! I've opened #277 to see if there's any others like this

Grundlefleck:
Excellent. Much appreciated.

PR #280, opened 2015-12-01
Create PowerShellEditorServices.yml

daviwil:
This change adds a project registration for the PowerShell Editor Services project.


shiftkey:
Thanks!

PR #293, opened 2015-12-20
updating some stuff

shiftkey:
This bumps all the dependencies to whatever is currently used on GitHub Pages.

I also tested the site when deployed to a fork - the main image needs to have it's relative path set, so I grabbed a helper method in 73908774b4495b2ee9364b6e0f26c1eb480b51d7 to do that.

There might be other areas that need this fix but whatever, you can see the live site here for yourself:

http://shiftkey.github.io/up-for-grabs.net/


shiftkey:
No heckling from the crowd on this, so I'm gonna merge it in and carry on...

Issue #288, opened 2015-12-11
Markdown-ify the project description


shiftkey:
Not sure many projects do this, but I think it's a nice bit of [flair](https://www.google.com/search?q=flair+office+space):

<img width="718" alt="screen shot 2015-12-12 at 9 18 43 am" src="https://cloud.githubusercontent.com/assets/359239/11757464/753c1316-a0b1-11e5-994c-9fb74e0d8911.png">

shiftkey:
Fixed in #301

PR #301, opened 2016-01-03
Markdown-ify the project description

zuta:
Added Markdown support for the project description.

Used showdown lib to parse markdown to HTML: https://github.com/showdownjs/showdown

Ref: https://github.com/up-for-grabs/up-for-grabs.net/issues/288

P.S. I'm noob


zuta:
Here is a screenshot:

pr

shiftkey:
This seems pretty great, @ZuTa - thanks for taking a shot at it! I had a quick skim to see if there was a way for us to do this server-side (so we can avoid the client side dependency completely), but couldn't spot anything easy to use - Jekyll doesn't give us the right hooks to parse markdown inside a template, but perhaps I'm missing something?

Thoughts @dahlbyk @jrusbatch?

zuta:
@shiftkey haven't worked with Jekyll before, so decided to go with the approach I'm most comfortable with. 

Thanks for getting back to me so fast.

shiftkey:
Just noticed that GitVersion is already inlining a link - let's switch that over to Markdown while we're in here https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_data/projects/gitversion.yml#L2

zuta:
The result of switching over to Markdown.

Before:
git2

After:
git1

zuta:
ping

shiftkey:
This all seems fine - there's just a merge conflict with this branch that's preventing me merging it.

cc @dahlbyk @jrusbatch to get a :thumbsup: before merging it in

zuta:
@shiftkey fixed merge conflict. Should be good to go.

shiftkey:
Alright, I'm gonna take this in. Thanks @ZuTa!

dahlbyk:
Belated :+1:

PR #302, opened 2016-01-04
fix multi-line string formatting

shiftkey:
I tried importing the projects YAML into `yamljs` and hit some parsing errors...

From [some documentation](http://symfony.com/doc/current/components/yaml/yaml_format.html) about YAML:

> When a string contains line breaks, you can use the literal style, indicated by the pipe (`|`), to indicate that the string will span several lines. In literals, newlines are preserved:

```
|
  \/ /| |\/| |
  / / | |  | |__
```

> Alternatively, strings can be written with the folded style, denoted by `>`, where each line break is replaced by a space:

```
>
  This is a very long sentence
  that spans several lines in the YAML
  but which will be rendered as a string
  without carriage returns.
```

This addresses the problem entries - I'll also do a deploy to my fork to indicate everything is fine with the existing site...

### DIFF
--- a/_data/projects/albacore.yml
+++ b/_data/projects/albacore.yml
@@ -1,6 +1,7 @@
 name: albacore
-desc: Albacore is a professional quality suite of Rake tasks for building .NET or
-  Mono based systems.
+desc: >
+  Albacore is a professional quality suite of Rake tasks for
+  building .NET or Mono based systems.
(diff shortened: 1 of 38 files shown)


shiftkey:
Fork deploy is fine - aside from the broken image addressed in #293

PR #305, opened 2016-01-04
Add Firefox

reuben:
Add a Firefox entry pointing to Josh Matthews' Bugs Ahoy! tool to find mentored bugs.


shiftkey:
Cool site, thanks for sharing!

PR #307, opened 2016-01-04
Create gravtr.yml

gfabrizio:


gravtr.yml
```diff
@@ -0,0 +1,9 @@
+name: Gravtr
+desc: A simple package to generate a gravatar url
+site: https://github.com/gfabricio/gravtr
+tags:
```
shiftkey, line 4: Would it be worth adding a `gravatar` tag here?

Issue #240, opened 2015-09-15
Support for multiple issue labels?


hrj:
In [gngr](https://github.com/UprootLabs/gngr), for example, we have three labels:
- need-code
- need-advice
- need-triage

for different types of contributors. It would be nice to list them separately in up-for-grabs.

shiftkey:
Closing in favour of #344

PR #310, opened 2016-01-04
Create hapi.yml

geek:


_data/projects/hapi.yml
```diff
@@ -0,0 +1,10 @@
+name: hapi
+desc: A rich framework for building applications and services
+site: http://hapijs.com/
+tags:
+- JavaScript
+- node.js
+- API
+upforgrabs:
+  name: Help Wanted
+  link: http://hapijs.com/contribute
```
shiftkey: I'd like for this link to point directly to the list of issues suited for new contributors. The landing page is nice, but it doesn't seem to render for me currently:



The first "See all" link is really great - and I hadn't thought about using the issue search in that way. We've had some discussions about supporting more flexibility with projects that span multiple repositories (or multiple labels like in #240) - will think further on this...

geek:
@shiftkey updated, thanks for the feedback

shiftkey:
Thanks!

PR #318, opened 2016-01-05
Add Emscripten

kripken:
This adds Emscripten to the list of projects here, using our existing tag "good first bug".

Thanks!


shiftkey:
Thanks!

PR #312, opened 2016-01-04
Markdown-ify the project description

L1fescape:
This fixes #288 
screen shot 2016-01-04 at 10 52 51 am

PR #326, opened 2016-01-07
Added Julia Language

kshyatt:


tkelman:
+1. We also have a separate "up for grabs" label but if we have to pick one I think the "intro issue" would be the better one to advertise.

_data/projects/julia.yml
```diff
@@ -0,0 +1,9 @@
+name: Julia Language
+desc: Julia is a high-level, high-performance dynamic programming language for technical computing. https://www.julialang.org
+site: https://github.com/JuliaLang/julia
```
shiftkey, line 3: Entirely up to you whether you'd like this to point to the repository or the top-level site, but this is the "project name" link the user will click.

_data/projects/julia.yml
```diff
@@ -0,0 +1,9 @@
+name: Julia Language
+desc: Julia is a high-level, high-performance dynamic programming language for technical computing. https://www.julialang.org
```
shiftkey: We don't render the links currently (see #312 for us supporting markdown) but using `https` here triggers a certificate error (as it expects julialang.org but gets github.com)

kshyatt:
Hi @shiftkey, I've updated the PR. Hopefully this is ok now? Thanks for your tips!

shiftkey:
@kshyatt looks great to me - thanks!

PR #334, opened 2016-01-14
Add  elementary OS Website

lewisgoddard:


lewisgoddard:
Okay, I'm not too concerned about it. Odd that you don't have to do that with the URL though.

shiftkey:
Thanks!

PR #346, opened 2016-01-25
Create LocalSupport.yml

tansaku:
All welcome - come join our community of learners polishing their development skills and contributing to good causes:

http://agileventures.org


shiftkey:
Thanks!

PR #352, opened 2016-02-03
Adding momentjs.com (docs site)

mattjohnsonpint:
To complement #246 

BTW - Up for grabs has worked well with us so far.   I'm hoping this addition will let some low-hanging fruit be given to those just getting started with open-source and might be more comfortable working on docs than on code.

Thanks!


shiftkey:
Thank you!

PR #355, opened 2016-02-04
Correct the react-bootstrap site url

MattCain:
I noticed that the site url for react-bootstrap was wrong so i've corrected it. Also I added a bootstrap tag as a couple of other projects are using it already.


shiftkey:
@MattCain thanks!

Issue #244, opened 2015-10-03
Multiple repos for a project


adamralph:
E.g. for ConfigR I have up-for-grabs labels in two repos, so I'd like to have two links. I guess I could add a whole separate "ConfigR Samples" entry but it would be nicer if both links were somehow shown against the single ConfigR entry.

I.e.

https://github.com/config-r/config-r/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs

and

https://github.com/config-r/config-r-samples/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs

shiftkey:
I guess the only hassle here is turning this element:

``` yaml
upforgrabs:
  name: up-for-grabs
  link: https://github.com/config-r/config-r/labels/up-for-grabs
```

Into something more like a dictionary (I'm not even sure if this is valid YAML):

``` yaml
upforgrabs:
  - { project: core, name: up-for-grabs, link: https://github.com/config-r/config-r/labels/up-for-grabs }
  - { project: samples, name: up-for-grabs, link: https://github.com/config-r/config-r-samples/labels/up-for-grabs }
```

Will think about this some more (having a separate project is a workaround, but doesn't help for maintainability and associating these two things).

adamralph:
Yeah, I can't claim to have thought it through very far.

The syntax highlighting suggests that it isn't valid YAML :wink:

How about this:

``` YAML
upforgrabs:
  core:
    name: up-for-grabs
    link: https://github.com/config-r/config-r/labels/up-for-grabs
  samples:
    name: up-for-grabs
    link: https://github.com/config-r/config-r-samples/labels/up-for-grabs
```

shiftkey:
I'd like to confirm that the `upforgrabs` element remains enumerable before we go down this path (otherwise generating the site content is going to be a pain) and also setup some CI hooks to verify that a PR contains valid YAML (so others know their PRs are structured right, and we don't accidentally merge  a PR and break the site).

adamralph:
Given that the schema would change from...

```
upforgrabs:
  name:
  link:
```

to

```
upforgrabs:
  project1:
    ...
  project2:
    ...
```

... I guess you'd have to migrate all the current entries to the new schema. Otherwise, you'd have to special case the "project names" `name` and `link`. Alternatively, you could use a new element instead of `upforgrabs`.

shiftkey:
[...]

PR #366, opened 2016-02-18
Adding Hoodie

gr2m:
I hope I did this right :) We love new contributors and put a lot of efforts to maintain a list of starter issues at all times.

Question: Is it possible to set two labels? `starter` is the label we add to issues that are perfect for new contributors. `ready` is an additional label which means it's up for grab. Once we remove `ready` it’s assigned to someone.

### DIFF
--- a/_data/projects/Hoodie.yml
+++ b/_data/projects/Hoodie.yml
@@ -1,6 +1,6 @@
 name: Hoodie
 desc: Generic Backend for frontend Apps
-site: https://github.com/dahlbyk/up-for-grabs.net
+site: http://hood.ie
 tags:
 - javascript
 - web
--- a/_data/projects/Hoodie.yml
+++ b/_data/projects/Hoodie.yml
@@ -1,0 +1,11 @@
+name: Hoodie
+desc: Generic Backend for frontend Apps
+site: https://github.com/dahlbyk/up-for-grabs.net
+tags:
+- javascript
+- web
+- offlinefirst
+- nobackend
+upforgrabs:
+  name: starter
+  link: http://go.hood.ie/hoodie-starter-issues


_data/projects/Hoodie.yml
```diff
@@ -0,0 +1,11 @@
+name: Hoodie
+desc: Generic Backend for frontend Apps
+site: https://github.com/dahlbyk/up-for-grabs.net
```
shiftkey: Set this to your project's home page :grin:

_data/projects/Hoodie.yml
```diff
@@ -0,0 +1,11 @@
...
+tags:
+- javascript
+- web
+- offlinefirst
+- nobackend
+upforgrabs:
+  name: starter
+  link: http://go.hood.ie/hoodie-starter-issues
```
shiftkey, line 11: We've got discussions about tracking multiple repositories #244 and multiple labels #240 but this is a fine starting point

shiftkey:
@gr2m welcome! I added a couple of notes here - it's pretty close to merging!

> Is it possible to set two labels? `starter` is the label we add to issues that are perfect for new contributors. `ready` is an additional label which means it's up for grab. Once we remove `ready` it’s assigned to someone.

I linked to a couple of related issues above, but we haven't quite got to the point of proposing changes to the markup structure (and then putting this onto the site). If you've got some ideas for tackling these I'd love to hear from you in those discussions...

gr2m:
How about you would support this format for multiple tags:

```
name: starter, ready
```

With the comma-separated list you would not break the current format. 

Is the PR okay like this now?

shiftkey:
@gr2m currently on the site we use the label here, just below the project name:



The original intent here is for projects to have a label that represents issues which they think are good for new contributors - so even though the site is called Up For Grabs, we don't dictate that you use that label.

And the other big benefit here is that, if you're using a GitHub link to represent your issues we can call the API and get the issue count from the JSON response. By adding in more flexibility in the format, we lose the ability to work with raw data and explore more project statistics - which would be cool to do on the site itself.

Anyway, we haven't settled on anything just yet - I have a side-project exploring this over at https://github.com/shiftkey/up-for-grabs-api-demo/ - and supporting different project workflows is something I'll keep in mind...

shiftkey:
> Is the PR okay like this now?

LGTM. Thanks!

_data/projects/Hoodie.yml
gr2m, line 11: How about this URL: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+org%3Ahoodiehq+label%3Astarter+label%3Aready+ – does that work with your labels showing the number of open issues?

shiftkey, line 11: It won't display the issue count because of the `org:hoodiehq` parameter

PR #371, opened 2016-02-25
Delete nservicebus.yml

adamralph:
We at https://github.com/Particular had a re-think about this. Whilst we welcome contribution, we feel that marking items as up for grabs is inappropriate for our repos.


shiftkey:
:cry: :wave:

daveaglick:
@adamralph I'd be curious what the thinking was if you're able to share. Certainly don't want to stir anything up, and obviously up-for-grabs labeling should be opt-in, but genuinely interested in what the downsides were.

Is it that the project is too complex and too mission-critical for novice users to jump in with high enough quality contributions (nothing wrong with this - I wouldn't expect NASA to accept novice PRs for the space station either)?

SimonCropp:
@shiftkey 

For me the important thing is: Particular and NServiceBus is commercial open source and not free source.

Free source can and should encourage.

Commercial open source should enable but not encourage.

"Up for grabs steps" over the line from enabling to encouraging.

Having been on the receiving end of commercial open source companies implying I can fix a bug myself, my only response is "I think you are misunderstanding your role in this equation".

Lets go through the possible things we could put on up for grabs.

### Bugs

A simple no. We should never be encouraging the community to fix any bug. If a community member wants to be proactive and go to the effort of fixing a bug that is fine. But our default stance should be we will fix all the bugs because that is why you pay for ongoing support.

### Doco

A simple no. We are a commercial company who values quality documentation. We should not be implying the community should do it for free for us.

### Features

Assuming someone was to go to the effort of adding a feature for us they are most likely going to only be interested doing the code part. We would most likely need to do the doco, release notes, customer communications etc. So it is a false assumption that the majority of the work would be done by the community.

If we do place features up on up for grabs there is a certain timeliness that we would need to address those features if someone implemented them. Lets say someone implemented one of those features a few months prio to a major release. It might not be practical or us to mange the release of that feature till after the next major. I think it is implied, by a feature existing on up for grabs, that if implemented it would be deployed in a reasonably timely manner.

So, in summary, I agree there is some possible positives that could fall out of being on up for grabs. I think the possible positives are outweighed by the possible negatives and effort involved in curating the list and managing the interaction. 

## Summary

Dont take this a criticism of what you are trying to achieve here. I think it is a great idea and well implemented. I just think it doesnt quite mesh with our specific scenario

shiftkey:
@SimonCropp @adamralph yeah, I totally empathize with the re-evaluation here, and gathered it was something in the vein that you've outlined above. Nothing personal.

daveaglick:
@SimonCropp Thanks a lot for sharing and taking the time to write such a thoughtful answer. Commercial OSS occupies an interesting space, one that I personally don't have a lot of experience with. Everything you said made a lot of sense, and were mostly considerations that hadn't occurred to me.

hobbyexplorer:
@SimonCropp  Thank you for the detailed answer!

Issue #317, opened 2016-01-04
CI - warn against multi-line strings


shiftkey:
As found in #302, we should warn when a PR contains a file with a value that spans multiple lines. It'd need to look at the source text, rather than the parsed YAML.

But basically this is bad:

```
desc: Albacore is a professional quality suite of Rake tasks for building .NET or
   Mono based systems.
```

This is what you should expect to see:

```
desc: > 
   Albacore is a professional quality suite of Rake tasks for building .NET or
   Mono based systems.
```

And this is also a valid option (although we shouldn't be adding in newlines here):

```
desc: | 
   Albacore is a professional quality suite of Rake tasks for building .NET or
   Mono based systems.
```

mwksl:
@shiftkey I think the regex in 9da28d0 should help with what you're looking for!

mwksl:
Had a typo in my last PR 😮, 0c6236f fixed it!

subsidel:
Apologies, GitHub is showing all amended versions of that commit for some reason.
However, I have opened a PR for this issue: https://github.com/up-for-grabs/up-for-grabs.net/pull/1036

ritwik12:
@subsidel That looks great.

SeanKilleen:
FYI I believe this was closed via #1038 and thus this issue can be closed.

ritwik12:
@SeanKilleen Thanks for reminding :)

PR #378, opened 2016-03-05
Multi line

mwksl:
Fixed typo to solve issue #317, regex checks for sentences that are broken up by newline marks and errors.

### DIFF
--- a/scripts/cibuild.rb
+++ b/scripts/cibuild.rb
@@ -34,7 +34,7 @@ def verify_file (f)
 
     dotNetInQuote = contents.index("- .NET")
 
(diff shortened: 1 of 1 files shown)


scripts/cibuild.rb
```diff
@@ -34,9 +34,16 @@ def verify_file (f)
 
     dotNetInQuote = contents.index("- .NET")
 
+    multiLineEntry = contents =~ /(^\s?.*\n+$)/
```
shiftkey, line 37: This seems to be looking at the whole YAML file. Should this just be the value of `desc`? I think we have to do this before parsing...

shiftkey:
Closing this out as inactive.

PR #387, opened 2016-03-16
Create publiclab.yml

jywarren:
We also have a first-timers-only listing at https://github.com/publiclab/plots2/labels/first-timers-only if that's helpful.


shiftkey:
Thanks!

PR #396, opened 2016-04-29
Add Flarum

franzliedke:
Could we add this? :)

Thanks for the great project!


shiftkey:
@franzliedke I can't see why not? :grin: thanks!

PR #402, opened 2016-05-04
Johnny-Five Wants You! For help making robots

rwaldron:


dahlbyk:
You might want to add tags for IoT, arduino and/or robotics for discoverability?

rwaldron:
@dahlbyk ok, added a few tags

dahlbyk:
Thanks!

PR #407, opened 2016-05-17
add rang cpp to list

agauniyal:


_data/projects/rang.yml
```diff
@@ -0,0 +1,15 @@
...
+- c++
+- cpp
+- cmake
+- windows
+- linux
+- coverage
+- travis
+
+upforgrabs:
+  name: up-for-grabs
```
shiftkey: This value should be the label you're using for your project: `help wanted`

agauniyal: @shiftkey thanks and updated.

shiftkey:
@agauniyal thanks!

PR #420, opened 2016-06-16
Create Rubberduck.yml

Vogel612:
Adds https://github.com/rubberduck-vba/Rubberduck to the up-for-grabs projects


shiftkey:
@Vogel612 thanks!

PR #428, opened 2016-07-08
Updated help wanted label

RichardLitt:
The search for "difficult: easy" was not correct. This fixes the link to go for the "help wanted" label, making the pipeline work much better.


shiftkey:
@RichardLitt thanks!

PR #430, opened 2016-07-12
Add libvirt project

andreabolognani:
libvirt uses GitHub as a read-only mirror, and in particular
doesn't track issues there. Hence, the URL points to a wiki
page where the available "bite-sized tasks" are listed.

### DIFF
--- a/_data/projects/libvirt.yml
+++ b/_data/projects/libvirt.yml
@@ -1,0 +1,13 @@
+name: libvirt
+desc: The virtualization API
+site: https://libvirt.org/
+tags:
+- api
+- c
+- cross-platform
+- library
+- oss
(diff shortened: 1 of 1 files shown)


andreabolognani:
I'm pretty confident that

```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
An error occurred while installing activesupport (5.0.0), and Bundler cannot
continue.
Make sure that `gem install activesupport -v '5.0.0'` succeeds before bundling.
```

was not caused by me dropping a `.yml` file into the repository :)

shiftkey:
@andreabolognani I shall look into this when I have a chance! I'll take this in in the meantime

PR #431, opened 2016-07-12
adding Mobius project

skaarthik:
Mobius project repo - https://github.com/Microsoft/Mobius


shiftkey:
@skaarthik thanks! I think the build is going to fail so I'm just going to take this in...

skaarthik:
Thanks @shiftkey

PR #446, opened 2016-08-29
Update nuget client project to point to github

Rutix:
NuGet has fully moved to github as seen on http://nuget.codeplex.com/ but the up for grabs still links to codeplex. This should fix this :)


dahlbyk:
Thanks! :bow:

PR #453, opened 2016-09-11
Add PHP reboot magazine to up-for-grabs.net

kapilsharma:
Add PHP reboot magazine to up-for-grabs.net
phpreboot.com is open source website of PHP community in Pune, India and also host PHP Reboot magazine, a magazine with good articles about PHP and related technologies.


shiftkey:
@kapilsharma thanks!

PR #457, opened 2016-09-18
Add support for launching the site using Docker

jrusbatch:
This PR creates a `Dockerfile` and `docker-compose.yml`, enabling it to be easily launched in a Docker container. I tested this using [Docker for Mac](https://docs.docker.com/docker-for-mac/) and [Docker for Windows](https://docs.docker.com/docker-for-windows/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM starefossen/github-pages:onbuild
+
+RUN bundle install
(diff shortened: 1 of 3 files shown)


shiftkey:
@jrusbatch you magnificent bastard. Cheers!

shiftkey:
@jrusbatch for reference, I wrote up a proposal a few days ago that I need to share - it mentioned docker-ifying things to make it easier to run this up. This is a great start.

jrusbatch:
@shiftkey Awesome! Sounds like we were on the same page!

PR #458, opened 2016-09-18
Update references to dahlbyk/up-for-grabs.net

jrusbatch:
Update references to the repo's old address under @dahlbyk's account to refer to this org instead.


dahlbyk:
May dahlbyk/up-for-grabs.net rest in peace

PR #460, opened 2016-09-21
removing rmw project. I have enough help now :)

andy5995:


dahlbyk:
> I have enough help now

Good problem to have! :tada:

andy5995:
:) Yeah.

I may need some help for some upcoming projects so I'll definitely keep up-for-grabs in mind. Thanks!

PR #463, opened 2016-09-22
Create macalifa.yml

thecodrr:
I would like to add my project to `up-for-grabs.net`

### DIFF
--- a/_data/projects/macalifa.yml
+++ b/_data/projects/macalifa.yml
@@ -1,0 +1,11 @@
+name: Macalifa
+desc: A music player for UWP and Windows 10.
+site: https://github.com/theweavrs/Macalifa
+tags:
+- .NET
+- C#
+- UWP
+- .NET Core
+upforgrabs:
+  name: up-for-grabs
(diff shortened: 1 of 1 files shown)


_data/projects/macalifa.yml
```diff
@@ -0,0 +1,11 @@
+name: Macalifa
+desc: A music player for UWP and Windows 10.
+site: https://github.com/theweavrs/Macalifa
+tags:
+- .NET
```
shiftkey: There's a issue reported in the CI tests - this should be in quote marks.

shiftkey (changes requested):

_data/projects/macalifa.yml
thecodrr: I added the quote marks for .NET, .NET Core and C# tag. Is that okay?

shiftkey (approved):

_data/projects/macalifa.yml
shiftkey: @theweavrs :thumbsup:

thecodrr: Thank you!

PR #464, opened 2016-09-27
Add Alda to projects

daveyarwood:
up-for-grabs.net is a great idea. Cheers!

### DIFF
--- a/_data/projects/alda.yml
+++ b/_data/projects/alda.yml
@@ -1,0 +1,12 @@
+name: Alda
+desc: A general-purpose, beginner-friendly music programming language
+site: https://github.com/alda-lang/alda
+tags:
+- music
+- language
+- clojure
+- java
+- zeromq
+upforgrabs:
+  name: low-hanging fruit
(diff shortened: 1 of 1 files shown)


_data/projects/alda.yml
```diff
@@ -0,0 +1,12 @@
...
+desc: A general-purpose, beginner-friendly music programming language
+site: https://github.com/alda-lang/alda
+tags:
+- music
+- language
+- clojure
+- java
+- zeromq
+upforgrabs:
+  name: low-hanging fruit
```
shiftkey, line 11: Looking at the issue list, it seems like both this label or `up-for-grabs` are used. Feel free to use either one.

daveyarwood: I think our `up for grabs` label may have a slightly different meaning than the purpose of up-for-grabs. We use `up for grabs` to identify issues that are ready for action, but which may or may not be beginner-friendly. `low-hanging fruit` is our label for issues which should be easy™ and are suitable for new contributors.

shiftkey: @daveyarwood all good :thumbsup:

PR #470, opened 2016-10-04
Add Numbas to the list of projects

christianp:


_data/projects/Numbas.yml
```diff
@@ -0,0 +1,19 @@
...
+- web
+- HTML
+- CSS
+- html5
+- algebra
+- CAS
+- quiz
+upforgrabs:
+  name: up-for-grabs
+  link: https://github.com/numbas/Numbas/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs
```
shiftkey: For a bit of :lipstick: you can use `https://github.com/numbas/Numbas/issues/labels/up-for-grabs` here

shiftkey:
@christianp thanks!

christianp:
I've noticed that the Numbas listing has a question mark on the 'up-for-grabs' label instead of a count. Is the link URL wrong? I notice that some projects don't have the '/issues/' part.

shiftkey:
@christianp I'm not sure what is happening - the tooltip says it only calculates it for GitHub projects so perhaps there's a bug in [here](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/javascripts/main.js#L73-L76)

PR #472, opened 2016-10-06
accept urls of the form github.com/user/repo/issues/labels/label

christianp:
... for github labels.

The same URL without `issues/` in the middle is currently accepted, but adding that in also works - github redirects to `github/com/user/repo/issues?q=label%3label`

I noticed this in #470, and thought it would be better to accept more versions of the URL than to require one in particular.

In my opinion, it would be better to explicitly say we're giving a github label in the YAML by writing a line `github-label: <label>` rather than `link: github.com/<user>/<repo>/labels/<label>`, and then construct the URL from the given label.

### DIFF
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -63,8 +63,9 @@
   })(window);
 
   var issueCount = function(project) {
+
     var a = $(project).find('.label a')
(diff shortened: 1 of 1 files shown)


shiftkey:
@christianp thanks for looking into this!

Let's try it out!

shiftkey:
Looks good!

PR #473, opened 2016-10-07
added Blitzed IRC Trivia

andy5995:

PR #474, opened 2016-10-08
[WIP] add some automated feedback for the label name

shiftkey:
🌵 **DON'T MERGE JUST YET** 🌵 

This automates a check that I do when reviewing new projects - see #473 for an example:
- the `upforgrabs.link` value should be encoded and appear in the `upforgrabs.url`
- for cases where we know it matches up to a GitHub repository, we should check it
- it's not a hard-and-fast rule

I might change this to be a suggestion rather than fail the build, but I don't have the infrastructure for that right now. 

I also cleaned up the projects which were caught by this check.

### DIFF
--- a/scripts/cibuild.rb
+++ b/scripts/cibuild.rb
@@ -20,7 +20,7 @@ def check_folder
 end
 
 def valid_url? (url)
-   begin
+  begin
     uri = URI.parse(url)
     uri.kind_of?(URI::HTTP) || uri.kind_of?(URI::HTTPS)
   rescue URI::InvalidURIError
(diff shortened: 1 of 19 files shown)


shiftkey:
Putting this back to [WIP] while I mull this over

scripts/cibuild.rb
```diff
@@ -105,6 +118,33 @@ def verify_file (f)
       return [f, error]
     end
 
+    if github_link?(link) then
+
+      if link.index('https://github.com/issues?q=') then
```
dahlbyk, line 123: Not sure what other parameters, if any, are valid for `/issues`, but the `q=` here seems overly specific; anything at `/issues` can be ignored from GH repo link parsing.

shiftkey: Yeah, I'll relax this - I didn't get any false positives from this but who knows what the future holds?!?!?!

Issue #476, opened 2016-10-09
suggest adding a field for "Contributing Guidelines" link


andy5995:
I've had several people find my project through up-for-grabs, and they seem to be first-time contributors to open source projects. They usually usually post on an issue and ask where to start, and have no experience submitting patches or PRs. So I direct them to a page where I've outlined instructions for contributing, doing pull requests, and how to sync a fork with the remote upstream.

This is why I'd like to suggest a link next to each project listing that  users can click, that will take them directly to the project maintainers contribution guidelines page. Having a ticket be the first page a user new to the project sees is not very productive, in my opinion. He or she will always need more information before contributing, at the very least a visit to the home page. Most of the users who contacted me seemed not to have even visited the home page.

dahlbyk:
This sounds like a great idea! Supporting an optional `contribute` key or equivalent should be pretty simple to add. Bonus points if anyone wants to use the GitHub API to scan existing projects for links to fill in.

ghost:
I can do this one! Using the github API for that sounds fun too :)

ghost:
Okay it's done but the script only works for the few projects that have CONTRIBUTING.md in their default branch. May or may not be due to bug in my script. Let me know if you have any modifications that could be made to get more added.

https://github.com/up-for-grabs/up-for-grabs.net/pull/497

ghost:
@ckhatri @dahlbyk oh I didn't see that someone had already made one :(

PR #487, opened 2016-10-26
adds a contributor guideline link above the issues, as per suggestion…

ckhatri:
… at issue #476 

Added some css styling on font size, went with a font size of 14px.
The contributor guideline link will only show up if the .yml file has a contribute: link, so it's optional. 

I didn't use the GitHub Api for existing projects but that might be something I do later.

Hope everything checks out, let me know!

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -23,6 +23,9 @@
     </tr>
     <tr>
        <td class="details">
+       <% if (project.contribute) { %>
+            <span class="contribute"><a href="<%-project.contribute %>">Contributing Guidelines</a></span>
+           <% } %>
         <p class="label"><a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>"><%-project.upforgrabs.name %></a></p>
        </td>
        <td class="details">
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -23,10 +23,12 @@
     </tr>
     <tr>
        <td class="details">
+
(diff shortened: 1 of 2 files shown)


shiftkey:
@ckhatri i like where this is heading. I tested it out with a project to see how it would render:



Could we put this below the label? And maybe make it visually different to the label element?

ckhatri:
Yup! Will get that done this weekend.

ckhatri:
uh better late than never?

stylesheets/stylesheet.css
```diff
@@ -522,6 +522,11 @@ form {
   word-break: break-all;
 }
 
+.projects .contribute {
+  font-size: 14px;
+  word-break: break-all;
```
shiftkey, line 527: There's some weird trimming here:



Should this be `keep-all`?

_includes/scripts.html
```diff
@@ -23,7 +23,12 @@
     </tr>
     <tr>
        <td class="details">
+
         <p class="label"><a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>"><%-project.upforgrabs.name %></a></p>
+       <br>
+       <% if (project.contribute) { %>
```
shiftkey, line 29: Could you add this field to the example payload in the `README.md`?

shiftkey (changes requested):
This is close - at this point I'm just thinking about whether we can style that better.

shiftkey:
Closing this out due to inactivity.

@ckhatri if you're still interested in implementing this, please open a fresh pull request

PR #491, opened 2016-10-30
tweak exception handling

shiftkey:
This makes it easier to find issues with your YAML syntax


shiftkey:
Example output:

```
> ruby scripts/cibuild.rb
301 files processed - 1 errors found:
/Users/shiftkey/src/up-for-grabs.net/_data/projects/akka.net.yml - Unknown exception for file: no implicit conversion of String into Integer
```

jrusbatch:
Awesome!

PR #494, opened 2016-11-07
Update debugger.html repo

Zacqary:
Updates the repo for Mozilla debugger.html. Current URL is no longer pointing to the official project repo, so the issue counter is out of date.


shiftkey:
@Zacqary thanks!

PR #496, opened 2016-11-09
Add Solr Express as a project

diegobrum:
**Solr Express**
A simple and lightweight query .NET library for Solr, in a controlled, buildable and fail fast way.


shiftkey:
@diegobrum thanks!

PR #497, opened 2016-11-11
Add contributing tag and utility script and render it for each project that has it 

### LINKED ISSUE #476: suggest adding a field for "Contributing Guidelines" link

ghost:
Add github_api gem to Gemfile so we can get contributing links.

Remove Gemfile.lock from repository and added to .gitignore. Not needed, is created by bundler.

Add find_contributing.rb to scripts. Apparently only works for like four projects - either due to an unknown bug or most must use other files or sections in their README.md... could still be useful with some modifications

Closes #476


shiftkey:
Closing this out as I'm unclear on what exactly we were planning to do here. Apologies.

PR #513, opened 2016-12-16
Create mediawiki.yml

waldyrious:


suriyaa:
👍

_data/projects/mediawiki.yml
```diff
@@ -0,0 +1,12 @@
...
+site: https://www.mediawiki.org
+tags:
+- mediawiki
+- wiki
+- web
+- cms
+- php
+upforgrabs:
+  name: Easy
+  link: https://phabricator.wikimedia.org/tag/easy/
```
shiftkey, line 12: Would it be better to navigate the user straight to the task list? e.g. https://phabricator.wikimedia.org/maniphest/query/PaVmrjYb6zFl/#R

waldyrious: I thought about that (I didn't like linking to a landing page either), but I'm not sure how permanent that link is. Should I change?

shiftkey: This is fine - "Open tasks" is available in the sidebar, it's not like it's hard to find...

PR #519, opened 2016-12-29
Add Distri for grabs

ethan-dorta:
Distri is two programs that work together for distributed computing, the server written in Node.js, and the client written in JavaScript, bringing the power of distributed computing to the web!


shiftkey:
@Flarp thanks!

PR #523, opened 2017-01-13
Add trae to projects

gillchristian:
This PR adds [`trae`](https://github.com/huemul/trae) to the projects.

### DIFF
--- a/_data/projects/trae.yml
+++ b/_data/projects/trae.yml
@@ -1,0 +1,12 @@
+name: trae
+desc: Minimalistic Fetch based HTTP client for the browser
+site: https://github.com/Huemul/trae
+tags:
+- oss
+- web
+- fetch
+- http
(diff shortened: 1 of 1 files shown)


_data/projects/trae.yml
```diff
@@ -0,0 +1,12 @@
...
+site: https://github.com/Huemul/trae
+tags:
+- oss
+- web
+- fetch
+- http
+- javascript
+upforgrabs:
+  name: up-for-grabs
+  link: https://github.com/Huemul/trae/labels/up-for-grabs
```
shiftkey, line 12: I can't see any issues created with this label. Would you be able to create those before I merge this in?

gillchristian: We had some but were taken already, I'll let you know when we have some so you can merge!

PR #525, opened 2017-01-13
Javascript not supported

chrilly1525:
This should hide the content of the page, then alert the user that JavaScript is not enabled. 

_Please comment and inform me on what I can do to improve my code._


shiftkey:
@chrilly1525 apologies for the delay in reviewing this. It's not clear to me how this gets detected when you run the site up, but I like the approach in #564 which renders the links server-side for those clients who don't have JS enabled.

PR #532, opened 2017-01-18
Add a couple of WindBot's projects

rmobis:
This adds a couple of the projects we manage at the WindBot organization.


shiftkey:
@rmobis thanks!

PR #533, opened 2017-01-21
Update and rename macalifa.yml to breadplayer.yml

thecodrr:
The project name and description changed. So, requesting changes here :)
Thanks.


shiftkey:
@theweavrs thanks!

Issue #512, opened 2016-12-16
Use project names in search, not just tags


waldyrious:
Otherwise we have to include the project name as a tag so that users can find them, which is kind of a hack.

shiftkey:
This might be a silly question from me, but if you already know the name of the project why wouldn't you go straight to Google/GitHub to find it?

waldyrious:
As a curious technically-minded person, or enthusiastic user but not yet involved in development, I might know about a project and want to get involved, but since they don't necessarily use the same ways to identify easy/beginner issues (different bug trackers, different label names, sometimes even manually curated lists rather than query results in an issue tracker, etc.), I would like to be able to use a resource like up-for-grabs to seek that information in a convenient way. Removing barriers is good for anyone, but especially important for newcomers :)

dahlbyk:
I wouldn't object to this, e.g. a React search could find the React project in addition to projects related to it.

ahuwaa:
Going to take a shot at this unless o.w. told not to

waldyrious:
@shiftkey with the merging of #534, I suppose this could be closed, right? I tried to test it but I'm hitting #594.

shiftkey:
@waldyrious I guess we can't close this out just yet. I've left a note in #534 to see if we can get to the bottom of it...

shiftkey:
Closing this out to get things off my plate.

@waldyrious if you would like to revisit this there's some details in #594 around what might have been the problem.

PR #534, opened 2017-01-21
add projects to searchable tags

ahuwaa:
Addresses #512 

- Adds project names as tags so they are searchable as well
- Does not double count projects with a tag matching the project name


shiftkey:
@ahuwaa I fixed up the merge conflict from the JS formatting and this seems good to go.

Thanks for contributing!

shiftkey:
@ahuwaa I had to revert this due to a runtime error reported in #594 - would you be able to look into this and see if we can get this integrated?

PR #536, opened 2017-01-22
Adds Wyam to projects

daveaglick:
Probably should have done this a while ago, but better late than never.


shiftkey:
@daveaglick thanks!

PR #542, opened 2017-01-28
Adding Twitter Cards meta-tags to the website

fronchetti:
I believe that just like the [Open Graph](http://ogp.me/), [Twitter Cards](https://dev.twitter.com/cards/overview) are also important for website advertising and statistical control. I added meta tags that I believe are relevant.

Note: I am studying about the repository and I believe I have done the right pull-request. However, I am a newcomer, please correct me if I did something wrong!


shiftkey:
@fronchetti thanks for getting started with this. Unfortunately due to how [Jekyll](http://jekyllrb.com) works, everything under `_site` is generated content - so these changes will be overwritten as soon as we update the `gh-pages` branch.

If you'd like to make changes to the site you'll need to look at these files:

 - [`_layouts/default.html`](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_layouts/default.html) - this is the structure of the page, and points out to a bunch of partial views
 - [`_includes/head.html`](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_includes/head.html) - this is where we include all the `meta` tags. I think this is where your changes need to go.

Feel free to keep pushing commits to this branch, or to open a fresh PR if you'd like a new start!

Issue #261, opened 2015-10-29
Github RateLimited


MaximRouiller:
So... each of us who connect to this page do the request. 

Each time. This lead to my IP being banned from the API as I was making too many of them. 

Of course, I could add my credentials but it wouldn't be very kosher. The API url doesn't take into account that I'm already logged in with Github. Maybe we could have a job that picks them up and generate a JSON file that you could load initially?

Issues doesn't need to be super accurate. A 24 hour update would be plenty IMHO.

Ideas?

shiftkey:
> Issues doesn't need to be super accurate. A 24 hour update would be plenty IMHO.

This should already be occurring: https://github.com/dahlbyk/up-for-grabs.net/blob/gh-pages/javascripts/main.js#L79 And we also defer fetching the issue count until you scroll the item into view

Can you show me in the Chrome Dev Tools what you're seeing after encountering being IP-banned?

MaximRouiller:
Screen cap:

![image](https://cloud.githubusercontent.com/assets/209384/10826166/c98069b2-7e3e-11e5-8d76-2b8208b8f138.png)

shiftkey:
@MaximRouiller sorry, I should have been clearer. What's the response payload from the GitHub API look like?

MaximRouiller:
Main issue is that I had use jspm without an account.

So if I explore the actual response from the server, I get:

``` javascript
{
  "message": "API rate limit exceeded for 204.101.154.3. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

MaximRouiller:
Corresponding HTTP Response Headers:

``` http
HTTP/1.1 403 Forbidden
Server: GitHub.com
Date: Thu, 29 Oct 2015 17:12:34 GMT
Content-Type: application/json; charset=utf-8
Transfer-Encoding: chunked
Status: 403 Forbidden
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 0
X-RateLimit-Reset: 1446139464
X-GitHub-Media-Type: github.v3
X-XSS-Protection: 1; mode=block
X-Frame-Options: deny
Content-Security-Policy: default-src 'none'
Access-Control-Allow-Credentials: true
[...]

PR #546, opened 2017-02-04
Use to github ETags to reduce rate limiting

bigwave:
#261 Use the Github [conditional request functionality](https://developer.github.com/v3/#conditional-requests) so that even when the local cache has expired you don't use some of the limited (60) requests allowed.

I'm not a JavaScript/JQuery programmer, so I'm happy to hear about any improvements to this.

### DIFF
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -82,21 +82,57 @@
         count.html(cached.count);
         return;
     }
-
-    $.ajax(url)
-      .done(function(data, textStatus, jqXHR) {
-        var resultCount = data && typeof data.length === 'number' ? data.length.toString() : '?';
-        count.html(resultCount);
-        storage.set(gh[1], { "count": resultCount, "date": new Date() });
-      })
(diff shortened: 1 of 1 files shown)


shiftkey:
@bigwave looks like an earlier PR around JS styling has triggered a conflict - are you interested in seeing if that can be resolved?

shiftkey:
@bigwave it looks like there's a merge conflict with the AJAX call - are you able to have a look at merge your changes into the existing code?

shiftkey:
Tagging `needs-work` as there are conflicts here to deal with. I'd like to see if we can get this in, but it's gonna take some time to be able to sit down and focus on resolving the conflicts and testing this.

shiftkey:
@bigwave thanks for opening this up, and apologies for not being able to land this earlier!

I've pulled in the important parts of this PR into #1276 so that we can wrap this logic in tests. If you have any additional thoughts to add, please comment there.

PR #564, opened 2017-03-01
render project links in index.html

christianp:
At the moment, the project data is held in JSON and the list is created
in javascript after the page loads. This breaks the site for users
without JS.

This moves the rendering of the project list to index.html, so it's
included in the static page, and then the renderProjects() function
shows/hides and reorders the projects as necessary.

If the JS doesn't run, you just see the complete list of projects in the
default order, without the filter interface.

This somewhat conflicts with #525, which assumes that the page is unusable if JS isn't supported.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -13,34 +13,6 @@
     <% }) %>
     </ul>
   </div>
-  <table class="projects">
-   <% _.each(projects, function(project){ %>
-    <tbody>
-    <tr>
-       <td colspan="2" class="title">
-           <span class="proj"><a href="<%-project.site %>"><%-project.name %></a></span>
-       </td>
-    </tr>
-    <tr>
-       <td class="details">
-        <p class="label"><a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>"><%-project.upforgrabs.name %></a></p>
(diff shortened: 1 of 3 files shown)


shiftkey:
@christianp this looks pretty great. The only regression I can notice here is that we no longer get the benefit of randomizing the project list on the client - [this code](https://github.com/up-for-grabs/up-for-grabs.net/blob/5823bfcad0645b7ab4c5537b9a5a6fc0cccefbe2/javascripts/projectsService.js#L92-L112) - and we're now doing thing alphabetically.

christianp:
@shiftkey is it really a regression? When the JS runs, the list is shuffled. I can't see how you'd shuffle the list per-client without JS, and on gh-pages.

dahlbyk:
It's a regression in that the top projects now get disproportionate exposure. If JS is available, we could suppress the HTML project list until we're able to shuffle it.

shiftkey:
Dropping this back to `need-work` because there are outstanding conflicts, and this changes the behaviour of the site to no longer randomize the list of projects.

shiftkey:
I'm going to close this out because #1412 is ready for discussion about a simpler first step to improve the experience for clients with JavaScript disabled.

PR #574, opened 2017-03-19
Modify up-for-grabs label for Algos

faheel:
Renamed the `up-for-grabs` label name to `Up for grabs` for the Algos project.


shiftkey:
@faheel thanks!

PR #506, opened 2016-12-09
adjust example code and overall wording

waldyrious:
- Reduce focus on second person (the person contributing a new project may not be the one running it -- we'd still want the contribution).
- Add examples for the upforgrabs label fields.
- Minor wording & punctuation tweaks.

PR #599, opened 2017-04-22
Fix link to "add-a-project"

pombredanne:
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


waldyrious:
Ah, that was my doing 😅 -- see #506.

shiftkey:
@pombredanne thanks!

Issue #607, opened 2017-04-30
issueCount doesn't work on "non-standard" GH issue URLs


aph3rson:
Right now, the `issueCount` function relies on the `link` attribute for each project. However, many of these links aren't in a standard format, and can't guarantee that additional options are being passed in this URL (both of which cause `issueCount` to not return a count).

Perhaps it might be better to update the function to compose its own URL to check against the API, by using the `site` and `up-for-grabs/name` attributes - that way, projects can link to a custom filter if need be, but the website will still render a count for viewers (albeit possibly inaccurately).

Related to #605.

aph3rson:
GitHub's Search API allows for 10x as many unauthenticated requests, as well as supporting searches for multiple labels (incl. negation), organizations, and exposing a single count in the JSON response (without dealing with pagination - returns up to 1000 responses in one query). It might be prudent to move to using this Search API, instead of using the current Issues API.

edit: x-ref https://github.com/up-for-grabs/up-for-grabs.net/issues/567#issuecomment-303563550

shiftkey:
This has been moved to cached numbers in the project data, so I'm going to close this out as resolved

PR #608, opened 2017-04-30
Fixing non-standardized URLs in project YMLs

### LINKED ISSUE #605: Multiple projects on GH don't render issue count
As mentioned above, multiple projects (such as Raiden Network, [raiden-network/raiden](https://github.com/raiden-network/raiden)) doesn't have a count rendered on the projects listing. Appears to be due to a malformed .yml, and not using standardized syntax.

See [raiden.yml](https://github.com/up-for-grabs/up-for-grabs.net/_data/projects/raiden.yml) as an example.

Tasks:
- [x] Update relevant YMLs
    - [x] Resin  
    - [x] cpp-ethereum  
    - [x] 8080py
    - [x] server for Node.js
    - [x] Golem
    - [x] Babel

See #607 for possibly preventing this from happening in the future.

aph3rson:
The patched YMLs were not able to be parsed by `issueCount`, which causes a "Issue count is only available for projects on GitHub" error. This changes the URL's format to an equivalent one, and restores the issue count for each project on the list.

See up-for-grabs/up-for-grabs.net#605 and up-for-grabs/up-for-grabs.net#607 for more information.
Closes #605.

### DIFF
--- a/_data/projects/8080py.yml
+++ b/_data/projects/8080py.yml
@@ -12,4 +12,4 @@ tags:
 - cli
 upforgrabs:
   name: bitesize
-  link: https://github.com/sadboyzvone/8080py/issues?q=is%3Aopen+is%3Aissue+label%3Abitesize
+  link: https://github.com/sadboyzvone/8080py/labels/bitesize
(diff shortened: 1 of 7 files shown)


shiftkey:
@aph3rson yep, this is easier than fixing our parsing code. Thanks for the investigating!

PR #612, opened 2017-05-04
Changing label for debugger.html project to be available

clarkbw:
We've received some feedback that "up for grabs" doesn't translate well beyond English so we've switched our local label to be `available`.  

Love this project, thanks for creating it.  Let us know how and if we can help.


shiftkey:
@clarkbw all good, and thanks for the update!

Issue #567, opened 2017-03-05
Incorrect GitHub Issue Count


dahlbyk:
https://twitter.com/ziki_cz/status/837867100848848898

> CoreFX shows 50 on badge,but has 538 https://github.com/dotnet/corefx/labels/up%20for%20grabs … bug?

Bug. No pagination. As far as @shiftkey and I can tell, the REST API doesn't expose open count per label, but GraphQL does:

```graphql
{
  repository(owner: "dotnet", name: "corefx") {
    issues(labels: ["up for grabs"], states: [OPEN]) {
      totalCount // open count
    }
    label(name: "up for grabs") {
      issues {
        totalCount // total count...useful?
      }
    }
  }
}
```

But GraphQL (preview) requires authentication. So we need a proxy of some sort. Or in the short-term, we could just show 50+... I assume there's an HTTP header we could check to tell if there's a next page?

shiftkey:
> I assume there's an HTTP header we could check to tell if there's a next page?

It's a hack, but you can get an approximation within +/- a page using [the pagination headers](https://developer.github.com/v3/#pagination). The `last` link will give you the number of pages and the page size.

e.g. `<https://api.github.com/user/repos?page=50&per_page=100>; rel="last"`.

aph3rson:
I believe I've fixed this. As @shiftkey mentioned, the pagination headers mention if content has been paginated. Using some math, we can use the number of pages, as well as the number of items on the last page, to determine the actual issue count.

This has been submitted as PR #604.

aph3rson:
Just to give an update, #604 is still pending review, but I wanted to mention that a move to the Search API might be more prudent, as I mentioned in https://github.com/up-for-grabs/up-for-grabs.net/issues/607#issuecomment-300633183. In short, here's what's leading me to that conclusion:

* The results count is returned by the Search API, and doesn't require math to figure it out  
* Search API has more unauthenticated requests than the Issues API  
* Search API allows us to search entire organizations, multiple labels, counts on a milestone, etc.,  
* Related to above, submitters of projects can have a more granular control over what they want their count to be.

shiftkey:
I'm going to close this out as we don't have plans to use the GraphQL endpoint, but #1276 made an improvement here to show a better approximation (using the `Link` headers which indicate how many pages of results there are).

<img width="259" src="https://user-images.githubusercontent.com/359239/62941891-82070200-bdad-11e9-85ba-f332004eb6af.png">

[...]

PR #621, opened 2017-05-23
Adds The AMP Project as a project.

mrjoro:
Adds The AMP Project as a project.

/cc @cramforce

### DIFF
--- a/_data/projects/AMPProject.yml
+++ b/_data/projects/AMPProject.yml
@@ -1,0 +1,13 @@
+name: The AMP Project (Accelerated Mobile Pages)
+desc: The AMP Project provides a framework (built on standard web technologies) to make it easy to build fast web pages.
+site: https://github.com/ampproject/amphtml/
+tags:
+- amp
+- css
+- documentation
+- html
+- javascript
+- web
+upforgrabs:
+  name: Great First Issues
+  link: https://github.com/ampproject/amphtml/milestone/25
--- a/_data/projects/AMPProject.yml
+++ b/_data/projects/AMPProject.yml
@@ -9,5 +9,5 @@ tags:
 - javascript
 - web
 upforgrabs:
-  name: Great First Issues
-  link: https://github.com/ampproject/amphtml/milestone/25
+  name: Great First Issue
+  link: https://github.com/ampproject/amphtml/labels/Great%20First%20Issue
--- a/_data/projects/AMPProject.yml
+++ b/_data/projects/AMPProject.yml
@@ -1,5 +1,5 @@
 name: The AMP Project (Accelerated Mobile Pages)
-desc: The AMP Project provides a framework (built on standard web technologies) to make it easy to build fast web pages.
+desc: The AMP Project provides a web component library to make it easy to build fast web pages.
 site: https://github.com/ampproject/amphtml/
 tags:
 - amp


_data/projects/AMPProject.yml
```diff
@@ -0,0 +1,13 @@
...
+tags:
+- amp
+- css
+- documentation
+- html
+- javascript
+- web
+upforgrabs:
+  name: Great First Issues
+  link: https://github.com/ampproject/amphtml/milestone/25
```
aph3rson: FYI, the link you've used on L13 will not render an issue count on the website - it will only provide a link. If you have a label set up for GFIs, you can use that (in the form of `https://github.com/ampproject/amphtml/labels/label%20here` or similar). 

It may end up that we could do a count of issues on a certain milestone, but that's going to require an API shift (see https://github.com/up-for-grabs/up-for-grabs.net/issues/567#issuecomment-303563550). If we go that route, we'll do another issue/commit(s)/PR to clean up everything.

_data/projects/AMPProject.yml
```diff
@@ -0,0 +1,13 @@
+name: The AMP Project (Accelerated Mobile Pages)
+desc: The AMP Project provides a framework (built on standard web technologies) to make it easy to build fast web pages.
```
cramforce: I'd say "The AMP Project provides a web component library to make it easy to build fast web pages"

mrjoro: Done.

_data/projects/AMPProject.yml
mrjoro: Got it; will it render as a ? in that case?

We had debated about whether to use a label or milestone for these when we first created them, and ended up with a milestone for various reasons (e.g. we want to keep track of all of our bugs in a milestone).  We may revisit that if we find it's causing a lot of problems, but for now if it will render but without a count here I think that's okay.

aph3rson: > Got it; will it render as a ? in that case?

Correct, it will render as a `?`, and also (somewhat erroneously) tell users it can't get the issue count because it's not a Github project. The `?!` is for API errors (usu. seen with API rate limiting).

Your decision to put the issues under a milestone isn't unique, so that's totally fine. While I don't necessarily agree with the practice myself, it should form a link fine - that's passed directly into the `<a>` tag on the webpage, count or not. Should/when we move to the Search API (as I mentioned above), then we can make the change to grab the correct count and display it.

cramforce: @mrjoro We could just add a redundant label.

mrjoro: OK, done. :)  It does seem like labels are a more standard way of tracking these, and it shouldn't be too hard to maintain both since the rate of GFI creation is manageable.

_data/projects/AMPProject.yml
```diff
@@ -0,0 +1,13 @@
+name: The AMP Project (Accelerated Mobile Pages)
+desc: The AMP Project provides a web component library to make it easy to build fast web pages.
+site: https://github.com/ampproject/amphtml/
```
shiftkey, line 3: Suggestion: if the home page for the project is more interesting as an entry point, you can use that here. I'm going to merge this anyway, but feel free to submit another PR to tweak that based on how it looks on the site.

shiftkey:
@mrjoro thanks!

PR #641, opened 2017-06-22
Change label to `up-for-grabs`

drvinceknight:
We currently have `Easy_pull_request` but would prefer to avoid the word `Easy`.


shiftkey:
@drvinceknight thanks!

Issue #511, opened 2016-12-15
Visiting the site with javascript disabled does not display any projects


Hestben:
Hey!

I just wanted you to know, that if I visit the site ([link](http://up-for-grabs.net/#/)) with javascript disabled, only the static content is displayed, no projects are listed, and no message informing the user that the site requires javascript is displayed.

This could be useful, whenever a user of noscript visits the site.

shiftkey:
@Hestben good point. If someone wants to take a shot at this, I'd be happy to give some ideas.

arjunnn:
Hi, I want to help.  Please give me some pointers. Maybe a `<noscript>`?

waldyrious:
(Disclaimer: I haven't looked in detail at the current setup, I don't know much about using Jekyll)

Could we tell Jekyll to generate a static page listing all the contents of the _data/projects folder, and then progressively enhance it via javascript, to inject a search field and hide blocks whose tags ([or title](https://github.com/up-for-grabs/up-for-grabs.net/issues/512)) don't match the search query?

shiftkey:
A `<noscript>` tag on the home page would be a good first go at this.

As to statically generating the markup for each project, this is how we're generating the JSON for the projects (`jsonify` is included by default in Jekyll):

https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_includes/scripts.html#L47

The rest of this file is the templates that get rendered. I'm a bit hazy on how it all hangs together, due to it being so long since I've been in there, but I gather that's what you'd like to recreate as static markup.

You can have a read about how Jekyll manages data files here: http://jekyllrb.com/docs/datafiles/

waldyrious:
Thanks for the pointers! Unfortunately I can't promise I'll have the availability to work on this anytime soon, but hopefully others can :)

psanjay-1:
can i work on this issue?

0x006E:
Is this now also open? I want to contribute

conner-calhoun:
I added a noscript tag that displays a banner similar to the one found at the top of the page. How could I go about pushing my code?
![image](https://user-images.githubusercontent.com/18505623/30121016-702fd444-92f8-11e7-9100-cf0661e9df41.png)

shiftkey:
@callium if you can find the file in the repository that you need to update, you can [edit the file](https://help.github.com/articles/editing-files-in-another-user-s-repository/) in the web editor which will open a pull request for review.

[...]

PR #653, opened 2017-07-01
fix issue #511

mrtnmgs:
Fix issue #511 Visiting the site with javascript disabled does not display any projects.
Only the list displays (no filtering by tag). Also, the label doesn't display how many issues have this label (as the JS page does).
This is my first PR! Sorry if I'm doing anything wrong!


index.html
```diff
@@ -8,8 +8,39 @@
 <div class="block">
   <h1><span class="header-inside">Projects</span></h1>
   <div id="projects-panel">
+    <table class="projects">
```
aph3rson: Your `<table>` tags are outside of your `<noscript>` tags. As far as I can tell, this will cause issues - you should check to see if there's an empty set of `<table>`s if rendering the site with JS turned on.

index.html
```diff
@@ -8,8 +8,39 @@
...
+        {% assign project = p[1] %}
+        <tbody>
+          <tr>
+            <td colspan="2" class="title">
+              <span class="proj"><a href="{{project.site}}">{{project.name}}</a></span>
+            </td>
+          </tr>
+          <tr>
+            <td class="details">
+              <p class="label"><a href="{{ project.upforgrabs.link }}" title="View open issues for {{ project.name }}">{{ project.upforgrabs.name }}</a></p>
```
aph3rson: You mention open issues - however, it's guaranteed that that link will take you to a list of issues for that project. Some have added links to their contributors' guidelines, and some to a discussion board (iirc).

It might be better to generalize this verbiage, to something like "Contribute to XYZ project!"

index.html
```diff
@@ -8,8 +8,39 @@
...
+            </td>
+            <td class="details">
+              {% if project.desc %}
+                <span class="desc">{{ project.desc }}</span>
+              {% endif %}
+              {% if project.tags %}
+                <p class="tags">
+                  {% for tag in project.tags %}
+                    {{ tag }}{% if forloop.index != forloop.length %},{% endif %}
```
aph3rson: There should be a space after that comma, else you're going to see a list like this:  
tag1,tag2,tag3

For future reference, you can simplify that if statement like this:

```html
{{ tag }}{% if not forloop.last %}, {% endif %}
```

index.html
```diff
@@ -8,8 +8,39 @@
 <div class="block">
   <h1><span class="header-inside">Projects</span></h1>
   <div id="projects-panel">
+    <table class="projects">
+      <noscript>
+        {% for p in site.data.projects %}
+        {% assign project = p[1] %}
```
aph3rson: Just to check, do you need that array index? IIRC, `site.data.projects` is already an array of projects, and your forloop should be going through each project iteratively - what you've got looks like it's an array of arrays.

aph3rson (changes requested):

index.html
mrtnmgs: Good catch thanks!

index.html
mrtnmgs: This is how `site.data.projects` is structured – it doesn't work without line 14. There might be more elegant ways to iterate but I didn't find... I'm not really sure why the data is structured like this in the first place.

index.html
mrtnmgs: I reused the language from the js page

index.html
mrtnmgs: Thanks!

mrtnmgs: @aph3rson Do I have to do anything to add your changes to my commit? This is my first PR I feel a bit in the dark!

aph3rson: You'll just need to commit to the branch on your side of the PR. The PR will update automatically, and your commits will be shown here.

PR #654, opened 2017-07-01
Updated NUnit up-for-grabs label and tags

jnm2:
We recently switched to using `up-for-grabs` to indicate up-for-grabsness and would also like to catch up with some tags.

/cc @rprouse @CharliePoole @ChrisMaddock

### DIFF
--- a/_data/projects/nunit.yml
+++ b/_data/projects/nunit.yml
@@ -8,5 +8,5 @@ tags:
 - TDD
 
 upforgrabs:
-  name: easyfix
-  link: https://github.com/nunit/nunit/labels/easyfix
+  name: up-for-grabs
+  link: https://github.com/nunit/nunit/labels/up-for-grabs
--- a/_data/projects/nunit.yml
+++ b/_data/projects/nunit.yml
@@ -2,10 +2,17 @@ name: NUnit
 desc: NUnit is the most popular and widely used unit test framework for .NET
 site: https://github.com/nunit/nunit
 tags:
+- NUnit
 - ".NET"
+- ".NET Core"
+- dotnet
+- dotnetcore
+- cross-platform
 - C#
-- Testing
+- testing
+- unit testing
 - TDD
+- OSS
 
 upforgrabs:
   name: up-for-grabs
--- a/_data/projects/nunit.yml
+++ b/_data/projects/nunit.yml
@@ -5,8 +5,6 @@ tags:
 - NUnit
 - ".NET"
 - ".NET Core"
-- dotnet
-- dotnetcore
 - cross-platform
 - C#
 - testing


jnm2:
@rprouse Hmm, we aren't interested in updating our description or site right now by any chance, are we?

rprouse:
@jnm2 why do we need to update our site. I am not against it, but we are in the middle of the transition.

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jnm2:
No reason that I know, just checking because I didn't think of asking earlier. :+1:

aph3rson:
As per the CI note, please enclose the .NET-type tags in quotes. Otherwise, they're interpreted as a malformed YAML file.

See [coveralls.net's YAML](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_data/projects/coveralls.net.yml) for an example.

_data/projects/nunit.yml
```diff
@@ -2,10 +2,17 @@ name: NUnit
 desc: NUnit is the most popular and widely used unit test framework for .NET
 site: https://github.com/nunit/nunit
 tags:
-- ".NET"
+- NUnit
+- .NET
+- .NET Core
```
aph3rson: As per the CI failure notes, please be sure to enclose .NET-type tags with `"`s. Otherwise, they fail a YAML formatting check.

See [coveralls.net's YAML](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_data/projects/coveralls.net.yml) for an example.

aph3rson (changes requested):

_data/projects/nunit.yml
jnm2: Missed the failure, thanks! I guess I misread the YAML spec. 😛

aph3rson: No worries! Just keep an eye on the CI results - generally speaking, if a CI check fails, then your PR won't be approved here.

jnm2: Passed.

aph3rson: Awesome! You should be good to go - project adds usually get approved pretty quick.

jnm2:
(/cc @jrusbatch @shiftkey)

_data/projects/nunit.yml
```diff
@@ -2,11 +2,18 @@ name: NUnit
 desc: NUnit is the most popular and widely used unit test framework for .NET
 site: https://github.com/nunit/nunit
 tags:
+- NUnit
 - ".NET"
+- ".NET Core"
+- dotnet
+- dotnetcore
```
jrusbatch: `dotnet` and `dotnetcore` are a little redundant here, would you mind removing them? The `.NET` and `.NET Core` tags are already used by other projects.

jrusbatch (changes requested):

_data/projects/nunit.yml
jnm2: Sure, done.

jrusbatch (approved):

jrusbatch:
@jnm2 Thanks!

jnm2:
Thank you!

PR #668, opened 2017-07-14
Add MySqlConnector

bgrainger:
Add "up for grabs" issues from [MySqlConnector](https://github.com/mysql-net/MySqlConnector).


shiftkey:
@bgrainger thanks!

PR #679, opened 2017-07-31
Added Wikivoyage listings

nicolas-raoul:


shiftkey:
@nicolas-raoul thanks!

Issue #682, opened 2017-08-03
External links load within the same tab.


ifeanyidavid:
The external links to projects load in the same tab as the site thereby causing a user to leave the site without returning.

I modified the links by adding target="_blank" to allow external pages load in separate tabs.

shiftkey:
@ifeanyidavid that linked commit seems like a good fix, feel free to open a PR against the project so I can pull it in!

shiftkey:
LOL, just saw #683. Nevermind!

shiftkey:
Fixed by #683

PR #683, opened 2017-08-03
#682 - Added functionality to allow external project links open in a separate tab

ifeanyidavid:
Links to external projects currently open in the same tab and as a result, a user could wander off without returning.

I added a target="_blank" attribute to the links to let them open in a separate tab.

Cheers.


shiftkey:
@ifeanyidavid thanks for this!

ifeanyidavid:
Glad I could help. Thank you.

PR #686, opened 2017-08-07
Adds git-secret

sobolevn:
`git-secret` is looking for additional maintainers: https://github.com/sobolevn/git-secret/issues/100


shiftkey:
@sobolevn thanks!

PR #688, opened 2017-08-09
add blockstack.js. fixed typo in blockstack browser entry

larrysalibra:
Added an entry for the blockstack.js: https://github.com/blockstack/blockstack.js

I also fixed a typo in the blockstack-browser entry.

Awesome site @shiftkey !


shiftkey:
@larrysalibra thanks!

Issue #678, opened 2017-07-31
Clean up tags


nicolas-raoul:
Currently we have many similar tags:
```
2d
3d
abstraction
Active Directory
Actor
actormodel
AD
adobe
ADO.Net
ADO.NET
agplv3
ajax
akka
alamofire
album
algebra
algorithm
Algorithm
algorithms
Algorithms
amp
analysis
android
Android
"Android"
angular
Angular
"Angular"
Angular 2
angular-formly
angularjs
AngularJS
animated-gifs
Ant
antlr
apache
"Apache Cordova"
ApacheSpark
api
API
API 
apiary
apidocs
apollo
apple
appletv
Appointment
apps
AppSettings
ArcGIS
archiver
arduino
Arrays
Artificial Intelligence
ASM
asp.net
ASP.Net
ASP.NET
"asp.net 5"
aspnetcore
"ASP.NET Core"
ASP.NET Core
aspnetmvc
asp.net mvc
ASP.NET Web API
assembler
assembly
Assertions
assistant
A-Star
async
atom
Atom
atvjs
audience - novice
Audio
authentication
automation
Automation
Auto-mocking
aws
AWS
Azure
"Azure"
"Azure Service Bus"
"Azure Storage Queues"
Backdrop
BackdropCMS
Background
Bash
bcmath
bdd
BDD
beacons
beaglebone
Beautiful Soup
BFS
biblatex
bibtex
bignum
bindings
Bitbucket
ble
blender
blockchain
blog
BookSleeve
bootstrap
Bootstrap
bot
brackets
bridge
Brighter
broadcast
browser
 - browser
Browser
build
Build
builds
buildsuite
Business Intelligence
c
c#
c++
C
"C#"
"C++"
C#
C++
cache
Cache
"Caching"
Cairo
"Cake"
calendar
caliburn.micro
cannonjs
canvas
cartesian-product
CAS
cdn
chakra
chakracore
chat
chatbot
chocolatey
Chocolatey
ChocolateyGUI
ci
CI
Circuit Breaker
Civic Tech
cli
CLI
cli app
client
clientless
clojure
cloud
Cloud
cmake
cms
CMS
 - CMS
Cobalt - JVM Programming Language
code
code analysis
"code coverage"
code fix
coding
collaborate
collaboration
college
color
Color
colorimetry
Colorscheme
colour
combinations
combinatorics
Command Dispatcher
commandline
command-line
Command-line
Command Processor
commands
[...]

PR #694, opened 2017-08-16
Add Noda Time project data

jskeet:


_data/projects/nodatime.yml
```diff
@@ -0,0 +1,12 @@
+name: Noda Time
+desc: A better date/time API for .NET
+site: https://github.com/nodatime/nodatime
+tags:
+- C#
```
jskeet: Would .NET be useful here as well? It's not clear whether the .NET and C# tags are effectively equivalent...

shiftkey: We're mostly figuring out our tag set over in #678 but this is worth doing for the moment.

jskeet: Righto - added in another commit.

shiftkey:
@jskeet thanks!

PR #702, opened 2017-08-21
Add aptos project

jasonwalsh:
[`aptos`](https://github.com/pennsignals/aptos) is a tool for validating data using JSON Schema and converting JSON Schema documents into different data-interchange formats.

Verification steps:

    $ docker-compose up

Visit http://localhost:4000/


shiftkey:
@rightlag thanks!

Issue #666, opened 2017-07-12
Search


ritwik12:
Right now up-for-grabs have filter option only as "filter by tags".
I would suggest to include search by name.
Searching for the project on which we worked in the past or noticed as a good project but left it for future possibility is really hard.
Right now we need to go through the whole website to search for it.

Suppose I saw a project named "Thimble"
And kept it for future, But when I visit up-for-grabs again I need to go through all the projects for the one I wanted.

ritwik12:
#710 needs review to get merged, please review @shiftkey  and @dahlbyk .

ritwik12:
Merged #710

PR #710, opened 2017-08-28
Add filter by name

Motardo:
Hello,
I'm new to this project, and I thought I would make an attempt at a contribution. This is a proposed feature for #666. It adds an input/select box for the name of a project. The project list is filtered to show the match(es). Here is a screenshot of typing a project name:
selection_008

And a screenshot of two projects selected by name:
selection_007

Please suggest any changes or improvements. I will add some tests if this is the right approach.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -1,5 +1,13 @@
 <script type="text/html" id="projects-panel-template">
   <div class="filters-panel cf">
+    <h5>Filter by name: </h5>
+    <select class="names-filter" multiple data-placeholder="Select a project..." >
+    <% _.each(names, function(entry, key){ %>
+      <option value="<%-key%>"><%- entry.name%></option>
+    <% }) %>
+    </select>
+  </div>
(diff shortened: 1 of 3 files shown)


ritwik12:
@Motardo Is it able to retrieve names from the website or are you giving static data?

Motardo:
@ritwik12 It is getting the project names from the website. I basically just used the tags filter as a template and switched tags to names. The logic is even slightly simpler since each project only has one name.

ritwik12:
@Motardo  Alright, I suppose it is good to go then.

Motardo:
Awesome!
I saw some tests for the tags filter in [ProjectServiceSpec.js](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/tests/spec/ProjectsServiceSpec.js), and I was going add tests for the name filter, but I couldn't figure out how to even run the existing tests?

ritwik12:
Thanks for the contribution.

PR #717, opened 2017-09-06
Add Apache Impala

jbapple-cloudera:


shiftkey:
@jbapple-cloudera thanks!

PR #719, opened 2017-09-12
Rename Perspex to Avalonia

ForNeVeR:
Hello! It has been renamed for a while now, so I think we should also rename the project here, on up-for-grabs.


shiftkey:
@ForNeVeR thanks!

PR #721, opened 2017-09-12
Add pnpm project

zkochan:


shiftkey:
@zkochan oops, this should be `.yml` file like all the others - that's why CI is failing.

Everything else is :gem:

zkochan:
@shiftkey indeed. Fixed

shiftkey:
@zkochan thanks!

Issue #250, opened 2015-10-09
Filter should be an AND not an OR


FakeHaacked:
Hi all!

I was trying to combine filters and all it does in the search box is add _more_ results to the list. That seems to be not very useful. The reason I'm adding more filters is to narrow down the results, not expand them.

shiftkey:
:thumbsup:

shubheksha:
Any guidelines on what needs to be done to fix this?

shiftkey:
After a quick glance of projectsService.js, I think this is hiding in [applyTagsFilter](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/javascripts/projectsService.js#L2-L23) but I'm not 100% sure.

cc @chandu @dahlbyk for thoughts.

dahlbyk:
Yep, the [`projectNames`](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/javascripts/projectsService.js#L15-L18) list is built up as a distinct list of projects that "hit" on any of the tags. That variable can probably be discarded in favor of updating the `filter` clause below to check that the project includes all requested tags.

shiftkey:
:ok_hand:

chandu:
@dahlbyk @shiftkey Agreed. I will send a pull request with the required changes.

ritwik12:
PR #729 is able to solve this issue but the problem is that it works when you type a keyword and hit enter, but not when you type the keyword and then click the item from the drop-down.
@shiftkey  and @dahlbyk

erlend-sh:
I just ran into this today. I want to link newcomers in the Rust gamedev community to a tag combination like 
`https://up-for-grabs.net/#/tags/gamedev%2Crust`.

wolffles:
I think this can be closed its a duplicate, and the issue might already be fixed

erlend-sh:
Duplicate of which issue though? And if it’s solved, could you please share an example of how to do an AND filter?

wolffles:
the duplicate is here: https://github.com/up-for-grabs/up-for-grabs.net/issues/1115 
I was on the site yesterday, the filters are currently 'and'. They add upon each other already, if you select a label, then click on a popular tag it chooses all the labels with that tag... however whoever updated it missed the tags filter box which i then fixed in my pull request for the other issue. https://github.com/up-for-grabs/up-for-grabs.net/pull/1135

deltafactory:
[...]

PR #729, opened 2017-09-25
And filter for tags

clickr:
check out the working version at https://perthhackers.github.io/up-for-grabs.net/
sorry if I got the branching wrong, I'm new to this.
addressing issue #250

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-up-for-grabs.net
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -102,8 +102,24 @@
       tagsMap[tag.name.toLowerCase()] = tag;
     });
 
-    this.get = function (tags) {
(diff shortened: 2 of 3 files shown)


clickr:
Oh and I deleted the CNAME file to avoid conflicts with our repository

clickr:
Also I didn't quite understand all the legacy code so I did the best with what I know and left the rest alone.

cat-prices:
I noticed that it works when you type a keyword and hit enter, but not when you type the keyword and then click the item from the dropdown.

clickr:
I won't be able to look at this for a week or so. I guess the next step is to raise an issue and write a failing test. :-)

ritwik12:
@shiftkey I think we should close this. It's going to be more than two and a half years now and we haven't got any updates

ritwik12:
Closing this due to lack of activity from a long time.

PR #733, opened 2017-09-28
Add Godot Engine project

27thLiz:


27thLiz:
@shiftkey What does the "needs tasks" label mean? That the project doesn't currently have any up-for-grabs issues?
That's because we recently renamed that label from `junior job` to `hacktoberfest`.

How do you want me to resolve this, should I change the PR to use the `hacktoberfest` label instead, and then provide another PR when we change it back next month?

shiftkey:
@Hinsbart yeah, that was me tagging it to come back to later. I use it as a sanity check before merging to ensure some issues have been created.

> How do you want me to resolve this, should I change the PR to use the `hacktoberfest` label instead, and then provide another PR when we change it back next month?

That sounds great! Please update this PR to use that label.

27thLiz:
Sorry, I forgot about this.
Since hacktoberfest is over, the label changed back to its original name. So it should be fine now.

ritwik12:
@Hinsbart Thanks!!

PR #745, opened 2017-10-05
 after.html:changed _tag_ to _label_ 

andy5995:
_Label_ is much more specific


shiftkey:
@andy5995 great point! Thanks!

andy5995:
@shiftkey You're welcome!

PR #752, opened 2017-10-11
Remove Makesharp project - no longer maintained 

szymczakk:
As it is written in [readme of Makesharp](https://github.com/sapiens/MakeSharp/blob/master/README.md) this project is no longer maintained, so I think that it can be removed from the list.


shiftkey:
@szymczakk thanks!

PR #755, opened 2017-10-12
Rename "beginner" to "good first issue"

koppor:
I saw following popup on JabRef's GitHub repository:

grafik

Google did not find more about this movement, but I followed the suggestion by GitHub and renamed "beginner" to "good first issue". Thus, I also renamed it here in `jabref.yml` to ensure that the list of beginner issues is still correct.


shiftkey:
@koppor 👍

PR #757, opened 2017-10-12
Peek now uses "good first issue" tag

phw:
This seems to emerge as a standard on Github and is advertised by them.


shiftkey:
@phw thanks!

PR #762, opened 2017-10-12
Update allready.yml

BillWagner:
We're moving to the standard github labels for Help Wanted and Good First issue.


shiftkey:
@BillWagner thanks!

PR #764, opened 2017-10-13
Add Botwin project

jchannon:


_data/projects/Botwin.yml
```diff
@@ -0,0 +1,16 @@
+name: Botwin
+desc: Botwin is a library that allows Nancy-esque routing for use with ASP.Net Core.
+site: *https://github.com/jchannon/Botwin
```
shiftkey: Uh, I don't think this `*` needs to be here. But I don't think that's the error here 🤔

shiftkey:
@jchannon thanks!

jchannon:
@shiftkey thank you 👍

PR #765, opened 2017-10-13
ChakraCore: update to use label "help wanted"

dilijev:
See: https://github.com/Microsoft/ChakraCore/issues/3938

/cc @liminzhu


shiftkey:
@dilijev thanks!

PR #770, opened 2017-10-15
Add Node Cognitive Services

miparnisari:


miparnisari:
@ritwik12 what does that label mean? There are currently 4 pending tasks labeled as "good first issue".

shiftkey:
@miparnisari thanks!

PR #778, opened 2017-10-17
Numbas editor now uses "good first issue" label

christianp:
As recommended by GitHub


ritwik12:
@christianp Thanks :)

PR #785, opened 2017-10-23
mhwkb: update link, add tag, change label

andy5995:


_data/projects/mhwkb.yml
```diff
@@ -1,17 +1,18 @@
...
 - css
+- bootstrap
 - c
 - design
-- web
+- website
 - knowledge base
 - mentoring
 upforgrabs:
   name: up-for-grabs
```
shiftkey: This value here should match the label name: `good first issue`

shiftkey:
@andy5995 thanks!

PR #802, opened 2017-11-04
Add ddgr.yml.

jarun:
ddgr is a new project I am working on. It enables one to search DuckDuckGo from
the terminal.

I noticed a high demand for this utility in several discussions involving
googler, my other project to google from the terminal. Privacy is of the topmost
priority for many internet users and DuckDuckGo does a fine job at that.

It would be great to have some collaborators to work with and discuss. Some of
my utilities are already up-for-grabs. Can't think of a better place!


shiftkey:
@jarun thanks!

PR #831, opened 2017-11-30
Update ripme.yml to use "help wanted" tag (See RipMeApp/ripme#298)

metaprime:
https://github.com/RipMeApp/ripme/issues/298


shiftkey:
@metaprime thanks!

PR #832, opened 2017-12-01
Update shellrb.yml

faraazahmad:
Update the tag from help wanted to up for grabs

### DIFF
--- a/_data/projects/shellrb.yml
+++ b/_data/projects/shellrb.yml
@@ -8,5 +8,5 @@ tags:
 - Shell
 
 upforgrabs:
-  name: Help Wanted
+  name: Up For Grabs
   link: https://github.com/faraazahmad/shellrb/labels/help%20wanted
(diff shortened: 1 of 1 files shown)


_data/projects/shellrb.yml
```diff
@@ -8,5 +8,5 @@ tags:
 - Shell
 
 upforgrabs:
-  name: Help Wanted
+  name: Up For Grabs
   link: https://github.com/faraazahmad/shellrb/labels/help%20wanted
```
shiftkey: The "help wanted" label on the issue tracker still exists. Is this intentional, or should it also be renamed to "up for grabs"?

faraazahmad: I'm renaming all to up for grabs

shiftkey: 👍 then you'll need to change the `link` value below when that's been updated on the issue tracker

shiftkey:
@faraazahmad thanks!

PR #843, opened 2017-12-11
(Thallium) Not quite good for newbies anymore...

dead-claudia:
Too much major work going on, and I'm the only contributor.


ritwik12:
@isiahmeadows I think it will not be a good thing to remove it.As maybe someone may find something good from it.

shiftkey:
If the maintainer wants to remove their project from this list, that's their choice.

@isiahmeadows thanks for the update!

ritwik12:
@shiftkey ohk sure 😇.. I'll keep that in mind

PR #844, opened 2017-12-13
Add Liferea (Linux GTK/GNOME feed reader)

lwindolf:


_data/projects/liferea.yml
```diff
@@ -0,0 +1,14 @@
...
+- c
+- gtk
+- gnome
+- python
+- rss
+- atom
+- feed
+upforgrabs:
+  name: help-wanted
+link: https://github.com/lwindolf/liferea
```
shiftkey: This should be indented, and it should point to a list of suitable issues like `https://github.com/lwindolf/liferea/labels/help-wanted`

lwindolf: Thank you for the hint!

ritwik12:
@lwindolf Thanks :)

PR #868, opened 2018-01-20
Update consul project label name

bertocq:
We've changed the label from `beginner-friendly` on to `Good first issue`

### DIFF
--- a/_data/projects/consul.yml
+++ b/_data/projects/consul.yml
@@ -11,5 +11,5 @@ tags:
 - open-government
 - Manage topics
 upforgrabs:
-  name: beginner-friendly
-  link: https://github.com/consul/consul/labels/beginner-friendly
+  name: Good first issue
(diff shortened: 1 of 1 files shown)


shiftkey:
@bertocq thanks! There don't seem to be any issues open with that label - is that deliberate?

bertocq:
No @shiftkey is coincidental 🙈, but more like... we're still struggling to create easy to pick ones (well described, with a low required knowledge of the codebase...) and whenever we create one it gets usually picked up quickly by "usual" contributors :(. But its on our 2018 wishes to focus more on this to avoid such a close & small contributors circle :)

Should we wait to get some issues labeled with it before making the change? The old label was removed/renamed I think :(

shiftkey:
@bertocq given you're already listed on the site I won't hold this up. Just wanted to confirm.

PR #873, opened 2018-01-29
ZetaGlest:update info

andy5995:
* added some tags
* changed the link to the issues so that tickets from multiple repos
from within the ZetaGlest organization will be displayed.


ritwik12:
@andy5995  Thanks !!

andy5995:
Thank you!

PR #881, opened 2018-02-20
Add Hypothesis to the projects list

Zac-HD:
Hypothesis is a Python library for property-based testing, with [lots of tagged first-timer issues](https://github.com/HypothesisWorks/hypothesis-python/labels/good%20first%20issue) and a [long-standing commitment to mentoring](https://github.com/HypothesisWorks/hypothesis-python/blob/master/CONTRIBUTING.rst).

I've now seen up-for-grabs suggested multiple times as a place to get ideas for you first open-source contribution, and that means we want to be on the list :smile:


ritwik12:
@Zac-HD Thanks!!

PR #885, opened 2018-02-21
Accessibility Text Enhancement

lilliekhogg:
Hi All!
First of all I'd like to thank you for making such a helpful site, as a beginner contributor its been extremely helpful in finding useful projects to work on, get involved in the contribution community and be able to discuss changes with other users! 

Accessibility is always important on a website, so I have created this enhancement allowing those with site problems to zoom text in and out via the website. I do understand this can also be achieved in your web browser settings, but this would give a little extra accessibility to the up-for-grabs website. This is my first enhancement so any feedback, would be great help for future development on projects! The enhancement was creating using Javascript to increment and reduce the font size listed in the css file under the body tag.

Enhancement works as shown below:
pic 2
(Selecting the + button to increment the size of the body of text)
  
pic 6
(Selecting the - button to reduce the size of the body of text)

Many Thanks

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,6 +1,7 @@
 <header>
     <a id="forkme_banner" href="https://github.com/up-for-grabs/up-for-grabs.net">View on GitHub</a>
-    <div class="container">
+    <a class="incrementfont plus">A+</a> <a class="incrementfont minus">A-</a>  
+  <div class="container">
         <a id="go-back-home" href="/"><img src="{{ site.github.url }}/images/logo.png" alt="Up For Grabs"></a>
         <h1 class="ribbon">
             <strong class="ribbon-content">Explore open source projects and <a href="https://nikcodes.com/2013/05/10/new-contributor-jump-in">jump in!</a></strong>
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -70,8 +70,6 @@
   }
 </script>
 <script src="{{ site.github.url }}javascripts/lib/require.js" data-main="javascripts/app"></script>
-
-
 <script type="text/javascript">
   var gaJsHost = (("https:" == document.location.protocol) ? "https://ssl." : "http://www.");
   document.write(unescape("%3Cscript src='" + gaJsHost + "google-analytics.com/ga.js' type='text/javascript'%3E%3C/script%3E"));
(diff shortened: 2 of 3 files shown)


ritwik12:
@lilliekhogg We are happy that this site helped you.

ritwik12:
@dahlbyk  What do you say about this?

dahlbyk:
I'm on the fence here... most guidance ([1](https://www.levelaccess.com/font-resizing-guidance/), [2](https://ux.stackexchange.com/questions/703/font-size-amender), [3](http://simplyaccessible.com/article/resize-text/)) is several years old, and doesn't necessarily account for mobile considerations and the fact that all modern browsers zoom well.

From using https://lilliekhogg.github.io/up-for-grabs.net/#/, a few observations:

- Hovering over the A+ / A- should show an appropriate curser; currently it shows `cursor: text`
- If we're going to scale fonts, we should scale all fonts; currently some text doesn't change (e.g. tags)
- Line height is also not adjusting as the font size increases

While we can adjust how we write CSS, I expect these sorts of challenges are why folks these days tend to just lean on browser zoom. But I wasn't able to immediately track down any modern accessibility or UX guidance on this.

Thoughts?

ritwik12:
@lilliekhogg Also the A+ and A- should be visible at all times. I think it should be available at a corner even after scrolling and it should get invisible after some time and should come back when we move the cursor to it. Also, this is a good feature but alone it is a bit less in functionality. I would suggest you add something more to it.

savvaspc:
Hello guys. I'm thinking of doing some small stuff on this project and I wanted to start from something simple, so I thought this is a good start. I have already seen lilliekhogg's changes and made the font buttons have fixed position and show a hand cursor when hovered.

I thought it would be a good idea to make all font sizes relative, so that simply adjusting the body font-size in px will adjust all text on the page. I believe it's a better design in general. What do you think?

I'm currently looking into line-height and putting a hide option for the toolbar. Should I make a pull request with my ideas?

shiftkey:
@lilliekhogg apologies for the delay in getting to review this work. I updated this branch to build upon the latest work and resolve any conflicts, and we can view it in action here: https://deploy-preview-885--up-for-grabs-test-bench.netlify.com/

I'm aiming to look into it this week, but at first glance it looks very promising!

shiftkey:
I also see the comments in https://github.com/up-for-grabs/up-for-grabs.net/pull/885#issuecomment-387172586 and https://github.com/up-for-grabs/up-for-grabs.net/pull/885#issuecomment-387174542 haven't been talked about further.

Gonna think about this some more and see whether it's worth investing in this area (rather than leaning more on how browsers handle zoom).

shiftkey:
Closing this out due to inactivity, but if someone would like to open a fresh PR which addresses the feedback above we're happy to discuss further.

Thanks again @lilliekhogg for sharing your idea!

PR #882, opened 2018-02-21
pagination (#665)

djm158:
Went back and pulled in @layely 's code from #727 and made some modifications to hopefully get this working!

Issue #665, opened 2017-07-12
Pagination 


ritwik12:
This issue is related to #303 
I am opening this issue again as I think nobody is considering the old #303 
From my point of view, pagination is really required.
Whenever someone is searching for a good project to work on we always compare it with other projects.
Pagination helps us to see a certain list of projects at once.
It makes easy to search for the desired project.

aph3rson:
As per the initial issue, @apoorvnandan had a concept of this - realistically, either waiting for a PR from them and/or replicating their concept would do just fine.

cmdrwebdev:
Is this still open?

ritwik12:
@cmdrwebdev   Yes it is open and you may work and submit a PR.

ritwik12:
@cmdrwebdev  Have you tried?

layely:
@cmdrwebdev 
Is this still open ? I want to contribute.

ritwik12:
@cmdrwebdev Yes it is, As long as we don't get a PR.

layely:
@ritwik12 OK. So I am working on it

layely:
@ritwik12 

Here what i have:

- 20 projects per page 
- All projects are loaded at once of course
- when we move to the next page or previous page, auto-scroll (with animation) to the first project in that page 
- URI now looks like up-for-graps.net/#/page/<page-num>/tags/<tags> (eg. www.up-for-graps.net/#/page/3/ or www.up-for-graps.net/#/page/3/tags/.net

![up-for-graps-pagination](https://user-images.githubusercontent.com/19215803/30363252-02687414-984f-11e7-849f-5ac46a53727b.png)


![giphy](https://user-images.githubusercontent.com/19215803/30363891-7263172c-9851-11e7-984b-9646b0daa4a1.gif)

<div style="width:100%;height:0;padding-bottom:51%;position:relative;"><iframe src="https://giphy.com/embed/ZG6swaWPHDpJu" width="100%" height="100%" style="position:absolute" frameBorder="0" class="giphy-embed" allowFullScreen></iframe></div><p><a href="https://giphy.com/gifs/ZG6swaWPHDpJu">via GIPHY</a></p>

ritwik12:
@layely  Good work, Now please send a PR referring this issue.

layely:
@ritwik12 PR sent

ritwik12:
#727  is pending from about a month, please review @shiftkey and @dahlbyk

[...]

PR #887, opened 2018-02-23
Pagination gh-pages

djm158:
Making up for the hasty, late night, cowboy coded, half baked, merge conflict riddled PR that was #882. Hoping this will actually implement pagination as described in #665. Thanks to @layely for getting this started, I ended up cherry picking a lot of his code and doing some fine tuning on this. @ritwik12 if you can think of anyone to ping to review this please feel free!

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -9,7 +9,7 @@
   </div>
   <div class="filters-panel cf">
     <h5>Filter by tags: </h5>
-    <select class="tags-filter" multiple data-placeholder="Select a tag..." >
+    <select class="tags-filter" multiple data-placeholder="Select a tag...">
     <% _.each(tags, function(entry, key){ %>
       <option value="<%-key%>"><%- entry.name%>  (<%-entry.frequency%>)</option>
     <% }) %>
@@ -17,12 +17,17 @@
     <h5>Popular tags: </h5>
     <ul class="popular-tags">
     <% _.each(popularTags, function(entry, key){ %>
-      <li><a href="#/tags/<%-encodeURIComponent(entry.name)%>"><%-entry.name%></a> (<%-entry.frequency%>)</li>
+      <li><a href="#/page/1/tags/<%-encodeURIComponent(entry.name)%>"><%-entry.name%></a> (<%-entry.frequency%>)</li>
     <% }) %>
     </ul>
   </div>
   <table class="projects">
-   <% _.each(projects, function(project){ %>
+   <% _.each(projects, function(project, k){
+      var intervalStart = (currentPageNumber - 1 ) * nbrOfProjectsPerPage;
+      var intervalEnd =  intervalStart + nbrOfProjectsPerPage;
+
+   if(k >= intervalStart && k < intervalEnd) {
+    %>
     <tbody>
     <tr>
        <td colspan="2" class="title">
@@ -40,21 +45,37 @@
            <% if (project.tags) { %>
            <p class="tags">
                <% _.each(project.tags, function(tag, i) { %>
-                   <a href="#/tags/<%-encodeURIComponent(tag)%>"><%-tag%></a><%- i != project.tags.length-1 ? "," : "" %>
+                   <a href="#/page/1/tags/<%-encodeURIComponent(tag)%>"><%-tag%></a><%- i != project.tags.length-1 ? "," : "" %>
                <% }) %>
            </p>
-           <% } %>
+        <% }} %>
        </td>
     </tr>
     </tbody>
    <% }) %>
    </table>
+   <div class="center">
+     <div class="pagination">
+      <% var lastPage = parseInt((projects.length - 1) / nbrOfProjectsPerPage) + 1%>
+      <a id="previous" href="#/page/<%=currentPageNumber-1%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber <= 1) { %> class="not-active" <% } %> >&laquo; Prev</a>
+      <% _.each(_.range(1, projects.length + 1, nbrOfProjectsPerPage), function(i) {
+        var pageNumber = parseInt(i/nbrOfProjectsPerPage)+1;
+        if(pageNumber == 1 || pageNumber == lastPage || pageNumber == currentPageNumber) { %>
+        <a <%if(pageNumber == currentPageNumber) { %>
+          id="currentPage"
+          <% } %>
+          href="#/page/<%=pageNumber%>/tags/<%-encodeURIComponent(tagsString)%>"><%=pageNumber%></a>
+      <% }}) %>
+      <a id="next" href="#/page/<%=currentPageNumber-(-1)%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber >= lastPage) { %> class="not-active" <% } %>>Next &raquo;</a>
+    </div>
+  </div>
 </script>
 <script src="//cdnjs.cloudflare.com/ajax/libs/showdown/1.3.0/showdown.min.js"></script>
 <script type="text/javascript">
   var files = {{ site.data.projects | jsonify }};
   var projects = new Array();
   var converter = new showdown.Converter();
+  var nbrOfProjectsPerPage = 20;
   for (var fileName in files) {
     var project = files[fileName];
     project.desc = converter.makeHtml(project.desc);
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -57,7 +57,7 @@
    <div class="center">
      <div class="pagination">
       <% var lastPage = parseInt((projects.length - 1) / nbrOfProjectsPerPage) + 1%>
-      <a id="previous" href="#/page/<%=currentPageNumber-1%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber <= 1) { %> class="not-active" <% } %> >&laquo; Prev</a>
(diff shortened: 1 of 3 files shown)


ritwik12:
@djm158 I think @shiftkey  is the best person for this job.

djm158:
any update on this?

djm158:
https://djm158.github.io/up-for-grabs.net/#/

javascripts/main.js
```diff
@@ -30,27 +34,43 @@
...
+      var page = this.params["page"];
+      renderProjects("", "", page);
+      scrollUp();
     });
-
-    this.get("#/tags/:tags", function (context) {
-      var tags = (this.params["tags"] || "").toLowerCase().split(",");
-      renderProjects(tags);
+    
+    this.get("#/page/:page/tags/:tags", function (context) {
```
dahlbyk, line 61: Thoughts on flipping order on `page`/`tags` here? I think of this as navigating pages with `tags/:tags`, rather than the reverse.

javascripts/main.js
```diff
@@ -30,27 +34,43 @@
...
+      renderProjects("", "", page);
+      scrollUp();
+    })
+    this.get("#/page/:page/tags/", function (context) {
+      var page = this.params["page"];
+      renderProjects("", "", page);
+      scrollUp();
     });
-
-    this.get("#/tags/:tags", function (context) {
```
dahlbyk, line 42: Whether or not we flip `page`/`tags` in the paginated URL, we should keep `#/tags/:tags` working for folks who have shared deep links. It seems cleaner to suppress `page/1` rather than "redirect" to the `page/1` link.

javascripts/main.js
```diff
@@ -4,21 +4,25 @@
     compiledtemplateFn = null,
     projectsPanel = null;
 
-  var renderProjects = function (tags, names) {
+  var renderProjects = function (tags, names, page) {
```
dahlbyk, line 7: Would it be cleaner to have `renderProjects` accept `page` first? It seems like this would let us avoid some of the churn in the calls below.

Or, since route placeholders are consistent, we could refactor `renderProjects` to accept the Sammy `context` and just pull `tags`/`names`/`page` out of the routes in here. Honestly, that's probably cleanest - the route actions are essentially identical.

javascripts/main.js
```diff
@@ -30,27 +34,43 @@
     });
   };
 
+  //we scroll to the first project when user swich page
+  var scrollUp = function () {
+    $('html, body').animate({
+      scrollTop: $("#projects-panel").offset().top
```
dahlbyk, line 40: If one follows a direct link anywhere but the home page, they'll be scrolled directly to the middle of the page. Is that what we want?

It seems like we could reasonably skip the scroll if the current position is above `#projects-panel`, or we could switch the scroll trigger to be clicking on a link with `href^='#/'` (roughly)? Or is there a Sammy navigation event we could hook on?

javascripts/main.js
```diff
@@ -4,21 +4,25 @@
...
+      "currentPageNumber": page,
+      "tagsString": tagsString
     }));
 
     projectsPanel.find("select.tags-filter").chosen({
       no_results_text: "No tags found by that name.",
       width: "95%"
     }).val(tags).trigger('chosen:updated').change(function (e) {
-      window.location.href = "#/tags/" + encodeURIComponent(($(this).val() || ""));
+      window.location.href = "#/page/1/tags/" + encodeURIComponent(($(this).val() || ""));
```
dahlbyk, line 25: Assuming we fix the route, I'm inclined not to include `page/1` in links if we can help it.

_includes/scripts.html
```diff
@@ -40,21 +45,37 @@ <h5>Popular tags: </h5>
...
+      <a id="previous" href="#/page/<%=currentPageNumber-1%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber <= 1) { %> class="not-active" href="#/page/<%=currentPageNumber%>/tags/<%-encodeURIComponent(tagsString)%>" <% } %> >&laquo; Prev</a>
+      <% _.each(_.range(1, projects.length + 1, nbrOfProjectsPerPage), function(i) {
+        var pageNumber = parseInt(i/nbrOfProjectsPerPage)+1;
+        if(pageNumber == 1 || pageNumber == lastPage || pageNumber == currentPageNumber) { %>
+        <a <%if(pageNumber == currentPageNumber) { %>
+          id="currentPage"
+          <% } %>
+          href="#/page/<%=pageNumber%>/tags/<%-encodeURIComponent(tagsString)%>"><%=pageNumber%></a>
+      <% }}) %>
+      <a id="next" href="#/page/<%=currentPageNumber-(-1)%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber >= lastPage) { %> class="not-active" <% } %>>Next &raquo;</a>
```
dahlbyk: `currentPageNumber-(-1)` here fascinates me. Does `currentPageNumber+1` not work?

_includes/scripts.html
```diff
@@ -40,21 +45,37 @@ <h5>Popular tags: </h5>
...
-                   <a href="#/tags/<%-encodeURIComponent(tag)%>"><%-tag%></a><%- i != project.tags.length-1 ? "," : "" %>
+                   <a href="#/page/1/tags/<%-encodeURIComponent(tag)%>"><%-tag%></a><%- i != project.tags.length-1 ? "," : "" %>
                <% }) %>
            </p>
-           <% } %>
+        <% }} %>
        </td>
     </tr>
     </tbody>
    <% }) %>
```
dahlbyk: Should the new `}` on line 51 go here instead?

javascripts/main.js
```diff
@@ -4,21 +4,25 @@
     compiledtemplateFn = null,
     projectsPanel = null;
 
-  var renderProjects = function (tags, names) {
+  var renderProjects = function (tags, names, page) {
+    var tagsString = tags;
+    tags = (tags || "").toLowerCase().split(",")
```
dahlbyk, line 9: As far as I can tell, `selectedTags` is never actually used, so I propose aligning `tags` usage with `names`:

1. `tags` variable remains the `,`-delimited list.
2. `tags` split is deferred to [`projectsSvc.applyTagsFilter`](https://github.com/up-for-grabs/up-for-grabs.net/blob/8f2cabb8b9e86733b5f609351c28c82f9f3d0d81/javascripts/projectsService.js#L2-L5), which already handles it coming through un-split.
3. `selectedTags` can then be used instead of adding `tagsString`

You might check `blame` to make sure I'm not overlooking some reason why the `tags` split needed to happen where it did before you moved it here.

dahlbyk (changes requested):
@djm158 many belated thanks for this proposal! I've left a few thoughts.

Thanks @ritwik12 for the nudge!

ritwik12:
@dahlbyk Thanks a lot for your reviews.

ritwik12:
@djm158 I would suggest making changes so that we can merge this.

_includes/scripts.html
djm158: haha yes I would believe `currentPageNumber+1` would work 😄

javascripts/main.js
djm158: I think this idea makes sense!

djm158:
Working on the proposed changes, updates incoming!

_includes/scripts.html
djm158: actually after doing some digging it appears that `currentPageNumber+1` will actually evaluate to a string. i.e. if `currentPageNumber = '1'` then it will evaluate `currentPageNumber + 1` to `'11'`, but `currentPageNumber-(-1)` will evaluate to `2`.

djm158: i believe the correct thing to do would be `parseInt(currentPageNumber)+1`

_includes/scripts.html
djm158: I think so, since it's matching bracket is before the opening `<tbody>` tag.

_includes/scripts.html
```diff
@@ -66,7 +66,7 @@ <h5>Popular tags: </h5>
           <% } %>
           href="#/page/<%=pageNumber%>/tags/<%-encodeURIComponent(tagsString)%>"><%=pageNumber%></a>
       <% }}) %>
-      <a id="next" href="#/page/<%=currentPageNumber-(-1)%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber >= lastPage) { %> class="not-active" <% } %>>Next &raquo;</a>
+      <a id="next" href="#/page/<%=parseInt(currentPageNumber)+1%>/tags/<%-encodeURIComponent(tagsString)%>" <% if(currentPageNumber >= lastPage) { %> class="not-active" <% } %>>Next &raquo;</a>
```
dahlbyk: How about parsing `currentPageNumber` in `renderProjects`, so the template can be simpler? We allow number coersion above in the interval calculation, too; ideally we would replace anything that's not a positive page number with 1 before it reaches the template.

ritwik12:
@djm158 Hey, any updates?

djm158:
sorry all, was super busy and then went on vacation, will take a look again soon :)

ritwik12:
@djm158 please have a look if you get time.

ritwik12:
@shiftkey I think we should close this. It's going to be around two years now and we haven't got any updates

ritwik12:
Closing this due to lack of activity from a long time.

PR #903, opened 2018-03-25
Added Duality Docs Repo

ilexp:
Duality is a 2D game development framework that's [already part of upforgrabs](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_data/projects/duality.yml) for a few years. We recently split up our documentation, so it's now in a separate project, along with all issues related to the docs themselves. To get them listed again, I've added the docs project to upforgrabs separately, as we appreciate contributions to documentation as much as to the project core.

Let me know whether you think this makes sense for upforgrabs - not sure if tech writing should be a part of it, or whether the focus should be just coding itself, in which case just let me know and close the PR unmerged.

### DIFF
--- a/_data/projects/duality-docs.yml
+++ b/_data/projects/duality-docs.yml
@@ -1,0 +1,23 @@
+name: Duality Docs
+desc: >
+  A GitHub pages repo for the Duality projects developer manual. If you 
+  have a thing for technical writing and an affinity for game development, 
+  we could use your help!
+  Duality is a modular 2D game engine that provides its own visual editor. 
+  It's highly extensible, written entirely in C# and backed by OpenGL.
+site: http://dualitydocs.adamslair.net/
+tags:
+- markdown
+- documentation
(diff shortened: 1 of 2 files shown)


_data/projects/duality-docs.yml
```diff
@@ -0,0 +1,23 @@
+name: Duality Docs
+desc: >
```
ritwik12: @ilexp I think this is causing the problem in passing build for Travis CI, please refer other .yml in this repositories. You should have a single line description.

ilexp: Oh, I didn't realize this was failing the build. Sorry! Will look into this and update the PR.

ilexp:
Updated the description to be single-line, and while at it also did the same for the description of the already existing / merged duality.yml as well.

ritwik12:
@ilexp Thanks!!

PR #921, opened 2018-05-02
https://up-for-grabs.net/

dahlbyk:
I've updated DNS and such to enable https://up-for-grabs.net/, but it's not working due to mixed content warnings with `site.github.url`.

@shiftkey is there anyone internal you can poke about making `{{ site.github.url }}` resolve with `https` if the request came over `https`? I don't want to enable [Enforce HTTPS](https://help.github.com/articles/securing-your-github-pages-site-with-https/) until I can confirm everything is working as expected.


dahlbyk:
Oh hey, definitely merged this on accident. Luckily I just added a workaround from https://github.com/github/pages-gem/issues/238#issuecomment-385857100.

shiftkey:
> @shiftkey is there anyone internal you can poke about making `{{ site.github.url }}` resolve with https if the request came over `https`? I don't want to enable Enforce HTTPS until I can confirm everything is working as expected.

@dahlbyk is this still something you want me to ask about?

dahlbyk:
> @dahlbyk is this still something you want me to ask about?

The workaround shouldn't be necessary; a GH Pages request over https should reflect the same in its variables. I would recommend reopening https://github.com/github/pages-gem/issues/238 to track this.

PR #928, opened 2018-05-05
Create Quake-Info.yml

Kvaibhav01:


_data/projects/Quake-Info.yml
```diff
@@ -0,0 +1,14 @@
...
+- earthquake
+- earthquake-data
+- earthquake-api
+- json
+- list 
+- listView
+- USGS
+upforgrabs:
+  name: up-for-grabs
+  link: https://github.com/Kvaibhav01/Quake-Info/labels/up-for-grabs
```
ritwik12, line 14: @Kvaibhav01 It is better to have some issues related to this tag before merging so that others can contribute easily.

Kvaibhav01:
@ritwik12 What tasks are needed to be done?

_data/projects/Quake-Info.yml
shiftkey, line 14: @Kvaibhav01 it looks like you're using `help wanted` as the label.

If that's the case, `https://github.com/Kvaibhav01/Quake-Info/labels/help%20wanted` is the URL you should use here, and `'help wanted' for the `name` above.

PR #938, opened 2018-05-30
Fix Terasology's tag

syntaxi:
Terasology has changed to a different tag so I'm updating the reference on this site.


ritwik12:
@jellysnake Thanks!!

PR #953, opened 2018-06-20
Moving the project.yml to the correct location

Griffintaur:
Earlier I have created Periodic2048.yml at the wrong place (in the main directory) which should be in the projects directory.
Corrected it now.


ritwik12:
@Griffintaur Thanks!!

PR #971, opened 2018-07-19
Fix Docker-based dev environment

SantiagoTorres:
Hi, I've been wanting to add some projects that I lead (or know the leaders
of), but I noticed that deploying the current Docker-based dev environment
locally didn't quite work due to some compilation failures when building native
extensions to some of the dependencies.

I also found some error with the vendored contents:

```
  could not read file /usr/src/app/vendor/bundle/ruby/2.5.0/gems/jekyll-3.7.3/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb: Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document 'vendor/bundle/ruby/2.5.0/gems/jekyll-3.7.3/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the YAML front matter
```

I'm not quite a versed jekyll-ologist, but I found a fix
[here](https://github.com/jekyll/jekyll/issues/5267), which I assume translates
properly to this project (at least from what I could see after running
`docker-compose up`)

Hope this helps!

PR #972, opened 2018-07-19
add step to validate dockerfile

shiftkey:
This builds upon #971 and ensures we have a valid Docker setup as part of running CI tests


shiftkey:
I also upgraded to the latest `github-pages` gem and validated that the build (and Dockerfile) are still good.

PR #1014, opened 2018-10-03
Update contrast of footer text for a11y

theodoretan:
### Issue

Currently the colour contrast ratio of the footer text is below the standard of [WCAG colour contrast](https://www.w3.org/TR/WCAG21/#contrast-minimum) where it should be __at least__ `4.5:1`. As you can see below it is currently at `4.13:1` which is below the standard.

This is a problem for someone that is visually impaired and accessing this site where they might not be able to see the text correctly.

image
image

### Solution

I propose updating the contrast ratio up to `4.5:1` which is the minimum to satisfy WCAG AA requirements.

image
image

### DIFF
--- a/stylesheets/stylesheet.css
+++ b/stylesheets/stylesheet.css
@@ -221,7 +221,7 @@ footer {
   padding-bottom: 30px;
   text-align: center; }
   footer * {
-    color: #797979; }
+    color: #737373; }
 
 .container {
   *zoom: 1;


theodoretan:
ping @shiftkey @dahlbyk

shiftkey (approved):

shiftkey:
@theodoretan thanks for the contribution and the detailed writeup!

Issue #1015, opened 2018-10-03
formly-js/angular-formly has been archived


mquinson:
Hello,

I was lurking in the list and then I clicked on https://github.com/formly-js/angular-formly/labels/up-for-grabs 

This shows a tiny banner saying " This repository has been archived by the owner. It is now read-only". It sounds like it should be removed from your list, then. Newcomers may be disappointed if they realize too late that the project is not any more welcoming contributions.

Also, I was wondering whether there is a way to automatically detect when projects get archived, to automatically cure the list that you provide.

Thanks for this great initiative, I'll soon add some of my project to make my "beginner friendly" issues more visible.

ritwik12:
@mquinson Thanks for the info, appreciate your concern.

ritwik12:
@mquinson Your idea of automatically detecting archived projects is great and will be of great use. I am creating an issue for that so that it could be implemented.

ritwik12:
@mquinson merged

PR #1025, opened 2018-10-05
Delete angular-formly.yml as it is archived

ritwik12:
@shiftkey please review it and merge this if seems fine.


ritwik12:
This PR is related to #1015

PR #1036, opened 2018-10-09
Fix #317: Add warning for implicit multi-line strings 

### LINKED ISSUE #317: CI - warn against multi-line strings

subsidel:
At first I searched for a yaml linter, but the only ruby ones I found either wouldn't catch this issue or relied on now-deprecated gems.

This PR adds a minimal parser for yaml string lines, that is used if the values obtained by the real parser was a string, and works out if the original pre-parsed string from the file had a new line but did not have either of the `>` or `|` multi-line string tokens present.

This PR also fixes two yaml files that do not pass this new linting.

### DIFF
--- a/_data/projects/Virtual-Assistant.yml
+++ b/_data/projects/Virtual-Assistant.yml
@@ -1,6 +1,7 @@
 name: Virtual Assistant
-desc: A linux based Virtual assistant on Artificial Intelligence in C. 
-      Find restaurants in an area, Fetch weather information of any place, Google search, Youtube search, Play media files, Show calendar, etc.  
+desc: >
+  A linux based Virtual assistant on Artificial Intelligence in C. 
+  Find restaurants in an area, Fetch weather information of any place, Google search, Youtube search, Play media files, Show calendar, etc.  
 site: https://github.com/ritwik12/Virtual-Assistant
 tags:
 - C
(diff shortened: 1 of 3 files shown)


ritwik12:
@subsidel This looks great, One of the yml corrected by this script is mine only.

ritwik12:
@subsidel Sorry, I merged this by mistake. Looks good to me but will wait for @shiftkey to review this.

ritwik12:
Had to revert it. Please send PR for the same again. Sorry, for the trouble.

subsidel:
No worries, re-opened in https://github.com/up-for-grabs/up-for-grabs.net/pull/1038

PR #1048, opened 2018-10-13
Issue labels added

ramlaxman:


ramlaxman:
Hi @ritwik12,
can you show me steps how to delete commits from sent Pull request?
Thanks !

ritwik12:
@ramlaxman Commits in a PR are the commits that you make in your forked repo branch. I guess you haven't created a new branch and committed again and again even after sending PR. A good approach is to create a new branch for every PR you are going to make as once you have sent your PR for one thing then you can't send another until the previous one gets merged. That is why we use branches.

ritwik12:
@ramlaxman Now for the case if you have already created a mess and want to correct it.
First checkout to the branch, here as I can see your branch for this PR is `ramlaxman-patch-issue-label`
So,

`git checkout ramlaxman-patch-issue-label`

Then rebase your branch using `git rebase -i HEAD~n`, where n is the number of commits counting from the head. If you have created 2 commits and want to rebase those then use `git rebase -i HEAD~2`
After that you'll get a screen with your commits and in front of your commits "pick" will be written. If you want to remove those commits then you can write "drop" or "d" instead of "pick" in front of the commits you want to delete. Otherwise you can just remove the whole line and that commit will be lost.
  
If you want to remove a specific commit which is way far from the head and you don't know the count. Then you should use `git log` and look for the commit you want to correct. Once you find the commit then copy that commits commit id and go back. Then type  -
`git rebase -i "64f22dc6a6b5234f3458d03455266a9fdcf304ee"` This is the example how a commit id looks like. Then you will again get the same screen and you can continue further.

ritwik12:
@ramlaxman If you face any problem. You can ask.

ramlaxman:
Thanks @ritwik12 for advice. I am facing exactly same issues, will try out and let you know.

ritwik12:
@ramlaxman facing issues?

ramlaxman:
New links are not required and old ones are for promotion purpose.

PR #1044, opened 2018-10-11
add script to validate existing GitHub projects

shiftkey:
Fixes #1024 by using the GitHub API to see if the repository has been marked as archived:

```sh
$ GITHUB_ACCESS_TOKEN=[token] ruby ./scripts/validate-projects.rb
668 files processed - 11 errors found:
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/Distri-JS.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/jscs.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/room-sprawl.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/thinktecture-identityserver-access-token-validation.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/emberjs.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/restful-routing.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/frontkit.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/arcgis.pcl.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/opacapp.yml - The repository no longer exists on the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/thinktecture-identityserver.yml - Repository has been marked as archived through the GitHub API
/Users/shiftkey/Documents/GitHub/shiftkey-desktop/up-for-grabs.net/_data/projects/xamarin-forms-labs.yml - Repository has been marked as archived through the GitHub API
```

[...]

PR #1052, opened 2018-10-17
Cull archived projects

shiftkey:
Following on from #1044, this PR removes the found archived projects from our listing ☠️☠️☠️.

I re-ran the script and found that `_data/projects/adimportservice.yml` was recently marked as read-only, so it's already adding value to my workflow.

The second commit tidies up the output so that it only displays the path within the repository.


ritwik12:
@shiftkey That's a great work. Nice to have it here.

PR #1057, opened 2018-10-21
Deleted Webber.yml

dragon540:
I want to remove Webber from the platform since I am unable to contribute to it.


ritwik12:
@dragon540 Sure

PR #1071, opened 2018-10-25
Remove qpixel

ArtOfCode-:
QPixel hasn't been maintained for a while - to the extent that it's difficult to even get started any more :) It's not useful to have it on here any more, so I'm removing it.


ritwik12:
@ArtOfCode- Thanks!!

PR #1078, opened 2018-10-31
Fixed a subhead's cap

ariesti:
Made sure the subheads are of similar capitalization style


ritwik12:
@ariesti Thanks for corrections :)

PR #1086, opened 2018-11-08
metasmoke.yml: Clarify that this is for Stack Overflow / Stack Exchange

tripleee:
Clarify that this is a SO/SE-related project


shiftkey:
@tripleee thanks!

PR #1102, opened 2018-12-06
Delete mindfulness-at-the-computer.yml

SunyataZero:
We are pausing new contributions, so i'd like to remove our project from the site. Thank you for providing this service though, it's given people a chance to make their first contribution!


ritwik12:
@SunyataZero Thanks for having it here

PR #1104, opened 2018-12-08
Filter by name fix

Zarad1993:
### Problem
Filter by name does not seem to be working on the live site, and that is due to incorrect filtering.
 
See this GIF
issue-filter

### Solution

This PR Fixes this issue by sorting the project names first just how it's rendered by `scripts.html`
Then mapping the passed names into the right project.

### Proof

fix-to-filters

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -30,18 +30,20 @@
     names = _.map(names, function (entry) {
       return entry && entry.replace(/^\s+|\s+$/g, "");
     });
-    console.log(names);
-    console.log(projects[0]);
+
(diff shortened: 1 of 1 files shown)


ritwik12:
@nemequ That looks good and thanks a lot for pointing this out. I guess we missed it in the last PR.

ritwik12:
Fixes #1105

ritwik12:
@Zarad1993 Your implementation is case sensitive. Please have a look at #1109 and help us in correcting this.

Zarad1993:
@ritwik12 Sure thing. Please assign me to 1109. I'll get a PR out soon.

PR #1107, opened 2018-12-09
Fix: Alphabetical order in Filters

zeyadetman:


ritwik12:
@zeyadetman Had to revert as it was not working properly.

PR #1110, opened 2018-12-10
Filter by names case insensitive

### LINKED ISSUE #1109: Filters ordering is case sensitive
I noticed that the current "Filter by name:" is case sensitive and is ordering names Alphabetically but is also giving priority to Upper case.

Zarad1993:
**Problem**
Fixes #1109 

Names are being ordered in a case sensitive way.

**Solution**

This PR refactors the sort method to be used `getNames`, which ensures the names are always sorted when it's called and ensures it's rendered in an ordered case insensitive way.

Also tests two tests were added to Filter By Name to maintain stability

### Proof

fix-by-names

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -2,7 +2,7 @@
   <div class="filters-panel cf">
     <h5>Filter by name: </h5>
     <select class="names-filter" multiple data-placeholder="Select a project..." >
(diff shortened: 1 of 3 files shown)


Zarad1993:
@ritwik12 - Please check out this PR and let me know if anything more needs to be addressed. I'll be happy to help.

ritwik12:
@Zarad1993 That's great!! Thanks

Issue #1115, opened 2018-12-14
Improve Filters


ritwik12:
Right now we are able to apply all the filters simultaneously. The only problem is that, Once we apply a filter say "Filter by tags - Python"  and we again apply another filter say "Filter by name:" or "Filter by label:" then we should only be getting the names in the SELECT list to choose which are related to "python" or labels related to "python" only. But right now there is no effect on the SELECT after applying filter.

wanderdust:
Hey! I've been using the website, and the filter "Filter by tags" doesn't seem to be working. It only works when you use the popular tags ones. If nobody is doing this I would like to work on it. Thanks!

wolffles:
@wanderdust I already submitted a pull request fixing it. right above your comment.

BhaskarJoshi-01:
i want to take this issue

ritwik12:
Sure @LuciferJoshi, go on and let us know if you have some cool implementation to make things easier. Please do check what we already have

jordar411:
@ritwik12 Would you be opposed if we contributed a tag counter on each filter? There are so many different tags, and some of them return zero results when applied. Could be an easy, yet effective value-add.

jordar411:
@ritwik12 excuse me, I meant "labels" in my previous post.

ritwik12:
@jordar411 Hey, I could not understand that clearly. But if you mean quantifying the LABELS then I think we already have it under `Popular tags:`. If it's something else then can you please elaborate with an example, please?

jordar411:
> @jordar411 Hey, I could not understand that clearly. But if you mean quantifying the LABELS then I think we already have it under `Popular tags:`. If it's something else then can you please elaborate with an example, please?

@ritwik12 
I'm sorry! I've attached screenshots for reference. ![image](https://user-images.githubusercontent.com/38548084/98423686-744a3600-205d-11eb-9b99-e8fff1d1b349.png)![image](https://user-images.githubusercontent.com/38548084/98423702-7ad8ad80-205d-11eb-97a5-24d19488dd1a.png)

The total amount of items references by a specific label. If you feel that this is redundant, I can disregard. 😄

ritwik12:
@jordar411 I got it now, thanks for the explanation. I think it will not be suitable to show them up in the field itself. We can introduce something same as "Popular tags:" like "Popular Labels:" and show the Highest count values like 5-6. What do you think? How would you like to show the counts instead?

[...]

PR #1135, opened 2019-01-10
improving filter issue #1115

wolffles:
The filters build on each other already, I did notice however when using the tags filter with anything else it would break. fixed it. 
https://github.com/up-for-grabs/up-for-grabs.net/issues/1115


ritwik12:
@wolffles Thanks!!

PR #1146, opened 2019-01-22
remove rogue character to restore fork button

shiftkey:
The current website has an issue with rendering the "View on GitHub" button:

I traced it to this character in the stylesheet. Removing that and running it locally puts the "View on GitHub" link back in the top-right:


ritwik12:
@shiftkey Great!!

Issue #1142, opened 2019-01-21
Sort by Recently updated 


ritwik12:
As per @AllanOcelot in #1140, it will be good to have a sort by `recently updated` or `most active` for projects.

Zarad1993:
Hey @ritwik12 - Just breaking down this, since we don't have `timestamps` on the project information. I believe we will be tracking down by using GH APIs
Maybe `recently updated` by hitting each repository on GH to pull down the latest PRs merged for a given week/month and sort accordingly? Hence, just a popup question, are there any keys to be stored somewhere for GH interaction in the project?

ritwik12:
@Zarad1993 Yes, we need to hit GH for this for each repository  and collect the last updated time for a repo.
What we can do is to collect last updated time of each repo in a file and just sort that file based on recently updated. I guess mapping based on key for projects and their recently updated time is very much needed. Doing all this from scratch will not be so easy so I will recommend using some pre built tool or API which can do this and then map projects accordingly.

ritwik12:
@Zarad1993 Since it could be a big feature request so it'll be good to break it down in steps so that help from others could also be taken if required.

ritwik12:
Few references for this: 
https://serverfault.com/questions/401437/how-to-retrieve-the-last-modification-date-of-all-files-in-a-git-repository

https://developer.github.com/v3/repos/commits/#list-commits-on-a-repository

https://github.com/google/go-github/issues/821

ritwik12:
@AllanOcelot It will be good to have your reviews on this issue time to time.

Zarad1993:
Thanks @ritwik12 . Just a note since we don't have specific key values for `repo` and `owner`
We do store the Site's github URL and that is just **assuming** the Sites that exist on GH and not other sources.

ritwik12:
@Zarad1993 Yeah, also not all site URL are github repo URLs but approx 80-90 percent URL are github URLs.

savvaspc:
Hello guys. I'm new here and new to contributing to OSS in general. I would like to suggest using the Github Rest API v3. A simple GET request can return all info about a repo, including "updated_at" and "pushed_at". For example, https://api.github.com/repos/up-for-grabs/up-for-grabs.net returns all the info about this repository.

savvaspc:
Update: It looks like Github API allows 60 requests/hour for unregistered requests. Using some form of OAuth can raise this to 5000/hour.

[...]

PR #1157, opened 2019-02-01
[WIP] Sorting projects based on "last_updated"

savvaspc:
I created a function in javascripts/projectsService.js that sorts the projects based on the "last_updated" field from Github API. For each project I check if it has a valid github repository, create a GET request in the form of "https://api.github.com/repos/{owner}/{repo}", and process it with AJAX. The timestamp is added as a new field in the project's object, and when all GET requests have been processed, they are sorted in descending order based on the timestamp. Projects that don't have a github repo are treated as if their timestamp is 0.

Of course this is not yet functional in the front end, but I believe it's a good first step. All I used is plain javascript and only a couple of functions from underscore.js.

Tries to implement #1142

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -264,9 +264,9 @@
       }
     }
   }
-  
+
   // dummy call to skip travis error "error  'sortProjectsByRecentlyUpdated' is defined but never used  no-unused-vars"
-  //TODO: Figure out where to put the sorted projects array and make a proper call to this function.
+  // TODO: Figure out where to put the sorted projects array and make a proper call to this function.
   sortProjectsByRecentlyUpdated(null);
 
   var ProjectsService = function(projectsData) {
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -187,8 +187,7 @@
         const stems = repo.split("/");
         const repoLocation = stems[stems.length-2] + "/" + stems[stems.length-1];
         return repoLocation;
-      }
-      else if (project.upforgrabs.link.includes("github.com")) {
+      } else if (project.upforgrabs.link.includes("github.com")) {
         repo = project.upforgrabs.link;
         repo = repo.substr(repo.indexOf("github.com")+11);
         const stems = repo.split("/");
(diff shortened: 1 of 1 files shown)


javascripts/projectsService.js
```diff
@@ -50,16 +45,13 @@
...
     // Make sure the names are sorted first. Then return the found index in the passed names
-    return _.filter(
-      _.map(projectNamesSorted, function(entry, key) {
-        if (names.indexOf(String(key)) > -1) {
-          return entry;
-        }
-      }),
-      function(entry) {
-        return entry || false;
+    return  _.filter(_.map(projectNamesSorted, function(entry, key) {
```
ritwik12: @savvas94 There are multiple spaces before "_", Also your PR have lots of other errors which are shown in travis logs. please have a look on those.

ritwik12 (changes requested):

savvaspc:
@ritwik12 All the travis errors have been fixed. I'm currently working on the html button part, but I need guidance on how to present the final data on screen. I guess I could figure out how to do it myself, but I don't wanna break the already established way of doing it.

- [x] Write function that sends GET Requests, processes responds, and finally, sorts projects.
- [ ] Add html element (probably some "Sort by most recent" button) and pair it with a js event that runs the above code
- [ ] Render the sorted array into html

ritwik12 (approved):

shiftkey (changes requested):
Holding this up for now because I feel this will only add to our pain in #261 and I'm not sure we can put together a client-side solution to that...

shiftkey:
@savvas94 thanks for your interest in contributing to this, and apologies for the delay in getting over to review this.

Because of our already significant rate limiting issues, I don't believe we should support fetching the `last_updated` field from the GitHub API. Because of that, I focused on what we could do server-side. I didn't really feel like maintaining infrastructure to handle this, so it's basically sat unaddressed until recently, when GitHub Actions gave me a way to run infrastructure-esque tasks easily.

I've opened #1329 to track what we can when generating the static site, and have a demo involved in #1346 which makes `last-updated` available when we generate the site, which I think is the best way to tackle this. I'm going to update #1346 to include a simple "last updated" tooltip to show this works, but you're more than welcome to take another shot at #1142 once that's reviewed and merged.

PR #1158, opened 2019-02-03
Create androidProjects.yml

anshulkhattar:


androidProjects.yml
```diff
@@ -0,0 +1,10 @@
+name: androidProjects
```
shiftkey, line 1: Could this be named something without reverting to camelcasing?

Something like `Beginner-Level-Android-Studio-Apps`?

androidProjects.yml
```diff
@@ -0,0 +1,10 @@
...
+desc: this repository consists of some basic projects of android studio
```
shiftkey, line 2: Suggested change: `A collection of basic Android projects that are tailored to new contributors`

androidProjects.yml
anshulkhattar: I'm unable to change the file, the edit option is blocked.

shiftkey: @anshulkhattar the fork has been deleted

shiftkey:
@anshulkhattar as we're unable to make changes as part of the code review, I'm going to close out this PR. Please open a fresh PR and keep the fork around so we can get the PR into a state where it's good to merge.

PR #1179, opened 2019-02-20
fixed the broken "popular tags"

johanforslund:
When clicking on one of the "Popular tags", the projects doesn't get filtered because the query parameter is the name instead of the 'ID'. This fixes that problem by searching for the project's ID before querying.


shiftkey:
@johanforslund apologies for the delay in replying to this. I had to merge in the current changes but I was able to retest on the staging URL and this looks good to me!

shiftkey (approved):

PR #1182, opened 2019-02-27
Eslint

andrewkoung:
Surprised I didn't see ESLint on the list, so I would like to add this project too.


shiftkey:
@andrewkoung thanks!

PR #1185, opened 2019-03-07
Cross-platform UI graphic design tool

FallenAngel97:
Here's my project, which aims to complete the gap for such platforms as Linux, which are lacking of support of tools, which can help create user interfaces


ritwik12:
@FallenAngel97 Thanks!!

FallenAngel97:
Thank you

PR #1189, opened 2019-03-15
Add two more Kentico repositories

Simply007:
* I have added two more kentico repositories
* Added Caas label to the all Kentico repos (Content as a sevice)


ritwik12:
@Simply007 Thanks!!

PR #1195, opened 2019-03-25
Update project Human Connection

roschaefer:
* Update slogan - we encourage active citizenship
* Add a new project description page
* Update list of technologies, we did a re-engineering so this list is quite different now.
* Update issue link as we moved to a monorepo


ritwik12:
@roschaefer Thanks!!

PR #1197, opened 2019-03-29
Add libertempo web and api files

prytoegrian:
Hi !

In this PR, I added two projects I'm working on. The code is still in a refactoring process, and the business behind is very interesting, so I think they could be ideals for beginners.

NB : Links in « popular tags » weren't displayed as such, so i forced a cursor pointer since it's better for UX.

Thanks !


ritwik12:
@prytoegrian Thanks!!

PR #1200, opened 2019-04-03
Add project PRoot

oxr463:
_chroot, mount --bind, and binfmt_misc without privilege/setup for Linux_

https://proot-me.github.io


shiftkey (approved):
@oxr463 thanks!

PR #1236, opened 2019-06-05
Delete discover-rewards-notifier.yml

nareddyt:
I no longer maintain this project. Just wanted to delete the project from this site so that first time contributors don't get confused.

Thanks again for accepting my project more than 1.5 years ago! I ended up helping quite a few first time contributors get their changes in 💯


ritwik12:
@nareddyt Thanks for having it here !!

PR #1250, opened 2019-06-23
changed tag/link for AncientBeast.com project

DreadKnight:
Got rid of the custom label and went with more standard stuff since GitHub started featuring them.


ritwik12:
@DreadKnight Thanks!!

PR #1257, opened 2019-06-26
add action for finding stale projects

shiftkey:
This PR adds a GitHub Action to enumerate the list of projects (using the GitHub API) and find any that are:

 - marked as archived by the repository owner
 - returning a 404 (so they can't be found where they were expected to be)

These should be cleaned up, so the other feature of this action is to create PRs to remove them. I originally had a script in the repository that would do this manually, but given Actions can be run on a schedule this felt like a great chance to automate the process.

You can see what this looks like over on my fork: https://github.com/shiftkey/up-for-grabs.net/pulls

The action won't bother creating a new PR if it finds an open PR that deletes the file.

Once this PR is merged, I'm going to follow up by wiring up with workflow file (which will run the action once a day) and we can verify it's working as expected here.


shiftkey:
@ritwik12 I'm going to merge this in early so I can wire it all up and test it out today, but please let me know if you have any questions about any of this!

shiftkey:
> Once this PR is merged, I'm going to follow up by wiring up with workflow file (which will run the action once a day) and we can verify it's working as expected here.

And I realise now that this organization hasn't been invited into the GitHub Actions beta. I should have checked earlier!



Once that's been enabled, I can actually complete this work and enable the cleanup action.

ritwik12:
@shiftkey This looks good, went through the code.

Issue #1259, opened 2019-06-26
Cleanup repo


ritwik12:
I saw lots of image files and other files lying around in the repo outside of directories, It will be better to keep them in directories to keep repo clean.

ritwik12:
While solving this issue better not to break things.

SeanKilleen:
@ritwik12 can you elaborate a little further on what else you'd like cleaned up? I've got the icons as part of #1274 but am not sure what your other goals might be for this.

ritwik12:
@SeanKilleen Yeah, thanks for that. I mean with every other file which might be lying here and there without any logic. This issue mainly concerns the better file structure or hierarchy to keep files of a project in open source. For eg: The same way we keep all CSS files in style named folder and main files in src folder for a website. Meaning to reduce as many files lying outside of directories as possible.

shiftkey:
@ritwik12 as it's not clear to me what else needs to be done, I'm going to close this out due to inactivity. Please open new issues if there is anything else that needs work here.

PR #1274, opened 2019-07-01
Move favicons into an icon folder

SeanKilleen:
Supports #1259 by moving icon files into their own `icons` folder, ensuring that we update the html & browserconfig files as well.


shiftkey (approved):

shiftkey:
Thanks @SeanKilleen!

PR #604, opened 2017-04-30
Add support for projects with 30+ active issues

aph3rson:
This aims to fix the issues seen in projects with a high number of relevant issues, by correctly interpreting Github's `Link` response header.

If a project's issue count is polled, this will listen use the `last` page reference in that header, along with some math ((last-1 * 30) + lastCount) to determine how many issues truly exist.

This should fix #567.

PR #1276, opened 2019-07-01
refactor to handle ETags and projects with large numbers of issues

shiftkey:
Related to #567 and #261

Supersedes #546 #604 

This PR extracts our "fetch issue count" which invokes `jQuery.ajax` into a testable module, to support these scenarios which have PRs open to add support:

 - #546 - Use to github ETags to reduce rate limiting
 - #604 - Add support for projects with 30+ active issues

I was able to reuse some of the code in this PR - particularly the logic - so thanks @bigwave and @aph3rson for opening those up.

### Why all this work?

The previous PRs currently have merge conflicts, and rather than trying to reapply them on top of the current `gh-pages` branch I took a step back to identify the current pain points for me as a maintainer:

 - the logic to fetch the issue count currently is complex and hard to follow
 - there are more scenarios that we want to support, but bolting them on here doesn't feel great
 - we have no tests in this area, so it's hard for us to know if a change will impact a scenario that was previously working

### The Approach

`jQuery.ajax` was first released in jQuery 1.5, way back in 2011. Since then modern browsers have standardized around the [`fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for making these requests. This API is promise-based, which is a bit more friendlier for writing tests with in Jest over callback-based APIs, and is easier to mock (I'm aware of [`mockjax`](https://github.com/jakerella/jquery-mockjax) but haven't used it in anger).

I broke this refactoring work up into three stages:

 1. sketch out an API to emulate the current behaviour
    - check for a project in local storage. if cached and not expired, return that count
    - store the updated values in local storage for next time
    - handle rate-limiting or other error cases
2. add in new scenarios and test cases
    - if cache has the `ETag` value from the last time it called the API, use that value when making a request to the GitHub API
    - if a project has more than 30 open issues, guess how many there are without making a second API call  (this will appear as a count of `N+` for some number N)
3. replace the old implementation with this code

The tests themselves are able to mock out the `fetch` API and `localStorage`, which helped to flesh out scenarios as well as catch bugs while I was doing the porting. I'll rework the tests to help ensure they make sense and are maintainable in the future.

### Polyfills

To ensure support for older browsers, there are two additional polyfills which are added to the project:

 - [`promise-polyfill`](https://github.com/taylorhakes/promise-polyfill) - to ensure the browser can work with promises
 - [`whatwg-fetch`](https://github.com/github/fetch) - to ensure a `window.fetch` API is available to the browser

### Remaining Work

 - [x] extract module and add some tests
 - [x] store rate-limit reset time in local storage, respect that if found
 - [x] identify other corner cases and wrap those in tests
 - [x] testing and more testing
 - [x] update project label widget to support three digit issue counts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 _site/
 node_modules/
 yarn-error.log
+coverage/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,35 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+      {
+        "type": "node",
+        "request": "launch",
+        "name": "Jest All",
+        "program": "${workspaceFolder}/node_modules/.bin/jest",
+        "args": ["--runInBand"],
+        "console": "integratedTerminal",
+        "internalConsoleOptions": "neverOpen",
+        "disableOptimisticBPs": true,
+        "windows": {
+          "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+        }
+      },
+      {
+        "type": "node",
+        "request": "launch",
+        "name": "Jest Current File",
+        "program": "${workspaceFolder}/node_modules/.bin/jest",
+        "args": [
+          "${fileBasenameNoExtension}",
+          "--config",
+          "jest.config.js"
+        ],
+        "console": "integratedTerminal",
+        "internalConsoleOptions": "neverOpen",
+        "disableOptimisticBPs": true,
+        "windows": {
+          "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+        }
+      }
+    ]
+  }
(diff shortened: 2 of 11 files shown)


ritwik12:
@shiftkey That's some good work, Thanks for this. Also, Thanks to @bigwave and @aph3rson for sending PR and helping us write this code and apologize for the delay.

shiftkey:
I think for the sake of being able to ship something and iterate on it I'd like to land this sometime this week. I added code coverage in this PR and you can see we're sitting at better than 90% for both the existing `projectsService` and this new `fetchIssueCount` module.

If we spot something weird after merging this PR, please open an issue with the details from the dev tools as well as details about your setup. This will help to identify if this is a new case we need to handle on our end.

shiftkey:
Feeling pretty confident about this PR based on testing the preview deploy just now. Here's what it looks like for projects with large numbers of available issues:



And the previous behaviour once you encounter the rate-limiting is also working as expected:



Unlike the previous implementation, the preview deploy no longer makes API requests when it knows it's been rate-limited, until that period has expired.

PR #1321, opened 2019-08-26
updating the upforgrabs to the correct label name / link

kentico-timothyf:
updating the upforgrabs to the correct label name / link


shiftkey (approved):
Thanks @kentico-timothyf!

PR #1328, opened 2019-09-02
reformat the YAML associated with the list of projects so it can round-trip without changes

shiftkey:
See https://github.com/up-for-grabs/up-for-grabs.net/issues/1329 for the big picture that's motivating this PR.

As part of enabling the ability to store statistics alongside projects (to help highlight to visitors which projects are active) I first need to be able to parse and update the YAML in a predictable way.

In advance of opening this action, this PR updates the current project set to be able to read and write the file without making changes, using the inbuilt YAML parsing that comes with Ruby.

 - [x] review outstanding files using `|` (keep new lines) in their description
 - [x] test deploy site and confirm projects render fine


_data/projects/30SecondsOfC++.yml
```diff
@@ -1,5 +1,7 @@
+---
```
shiftkey, line 1: This document separator is used in a few ways by YAML:

 - to prefix a version at the start of the file
 - to store several documents in the one file

Neither of these are used by us, but I couldn't find an option in [psych](https://github.com/ruby/psych) to turn this off.

_data/projects/30SecondsOfC++.yml
```diff
@@ -1,5 +1,7 @@
...
 name: 30 Seconds of C++
-desc: A collection of STL features (functions/libraries) of C++ which can be learned in 30 seconds or less.
+desc: >-
```
shiftkey, line 3: Round-tripping strings turned out to be a huge pain in this process, but I think I'm in a happy place with these two choices:

 - setting a line width of 100 characters when emitting the document means most lines are preserved without modification
 - For long `desc` fields, they are emitted with this `>-` which means "Replace newlines with spaces (folded)" and "No newline at end (strip)" - (citation: [yaml-multiline.info](https://yaml-multiline.info/))

Because we don't really support multiple paragraphs on the site itself, preserving strings in this way seems reasonable.

_data/projects/ACMESharp.yml
```diff
@@ -1,8 +1,8 @@
+---
 name: ACMESharp
-desc: .NET Library and PowerShell module for the ACME protocol used by Let's Encrypt
+desc: ".NET Library and PowerShell module for the ACME protocol used by Let's Encrypt"
```
shiftkey, line 3: I'm not 100% sure why a leading `.` here leads to such paranoid YAML generation,  but it seems like every string that starts with a `.` will be wrapped in quotes from now on.

_data/projects/ACMESharp.yml
```diff
@@ -1,8 +1,8 @@
...
 site: https://github.com/ebekker/ACMESharp
 tags:
-# Note those are tags categorizing your project, not issue labels.
```
shiftkey, line 5: By round-tripping the YAML we can clean up all these old comments from the template

_data/projects/Aidre.yml
```diff
@@ -1,5 +1,7 @@
+---
 name: Aidre
-desc: 🐍 📝 Python Translation Help Station - The software for 📄 .po file docs handling. A tool to help with official py docs translations
+desc: "\U0001F40D \U0001F4DD Python Translation Help Station - The software for \U0001F4C4 .po file docs
```
shiftkey, line 3: Emitting these YAML strings converted most of the emoji to their Unicode representation. Will confirm the client-side renders these fine on the staged deploy.

_data/projects/ReactStyleguidist.yml
```diff
@@ -9,4 +10,5 @@ tags:
 - UI
 upforgrabs:
   name: good first issue
-  link: https://github.com/styleguidist/react-styleguidist/issues?q=is%3Aissue+is%3Aopen+label%3A%22good%20first%20issue%22
+  link: >-
+    https://github.com/styleguidist/react-styleguidist/issues?q=is%3Aissue+is%3Aopen+label%3A%22good%20first%20issue%22
```
shiftkey, line 14: This change can be distilled down to a `label` URL, but I have work queued up to tackle these so I won't address it here

_data/projects/yes.yml
```diff
@@ -1,8 +1,8 @@
-name: yes
+---
+name: 'yes'
```
shiftkey, line 2: My favourite change as part of this PR was that `yes` (without the quotes) was being parsed and then emitted as a boolean `true`. Yay YAML :tada:

shiftkey (commented):
Unfortunately this PR touches all the project files, but I think that's worth doing now to simplify some upcoming PRs I'd like to open later this week. Here's the cliff notes about what's being changed currently.

_data/projects/Aidre.yml
shiftkey: Working as expected on the client:

shiftkey:
I'm merging this in early as it's a bunch of manual changes, and so that I can build up the more complex PRs to let them sit for longer this week.

PR #1346, opened 2019-09-23
caching stats in project output on build

### LINKED ISSUE #261: Github RateLimited

### LINKED ISSUE #1329: embedding stats into projects to avoid need to update GitHub API
I opened #1328 which is the first step towards tackling a long-time issue with the Up-for-Grabs website - **being able to show activity/stats on the site without needing to call the GitHub API from the user's browser**.

### Roadmap

The roadmap to do this consists of a few steps:

 - be able to update project files in a scripted way - #1328 
 - review project lists to identify what is currently stale - #1330 
 - script the steps to find and add stats to a project file - #1331 
 - update the site to look at these project files and skip accessing the GitHub API - #1346 
 - wire up scheduled task to refresh stats for project and open PR to update those which have changed - #1360 

### Why this approach? 

We're currently at 777 projects listed on the site, and I want to be able to simplify the maintenance work here in two areas:

 - automate as much work as possible, because humans are fallible
 - leverage the data we use as part of automating things

#1257 was my first attempt to leverage the GitHub API, and GitHub Actions, to find projects that can be removed because they're no longer active. But what if we went further than that?

What if we were able to read a project file, find the GitHub projects, and look at the label they listed? We could then:

 - confirm this label is still used on the project
 - see that there are open issues associated with this label
 - oh, and also look at the `updated_at` field for the repository while we're in here

It turns out you can write this as a single GraphQL query:

```gql
  query($owner: String!, $name: String!, $label: String!) {
    repository(owner: $owner, name: $name) {
      updatedAt
      label(name: $label) {
        issues(states: OPEN) {
          totalCount
        }
      }
    }
  }
```

Armed with this data, I think we can do some interesting things:

 - update the project files to embed these stats, which are then included whenever we publish the site through GitHub Pages, making it available to clients without them needing to call the GitHub API
 - as a scheduled task, refresh the stats and create a PR with any changed files
 - flag projects which no longer show the expected label for review by the maintainers

There's probably more ideas at play here, but for now I just wanted to brain dump where this is heading so that others can follow along.

### LINKED ISSUE #1369: Issue count not working
Issue count is shown `?!` instead of the total number of issues.

shiftkey:
Resolves #1329
Resolves #261 
Closes #1369

 - run the update stats action against all the projects
 - tweak the template for each project to look for this stats object and add the number

You can see this live at the Netlify deploy URL: https://deploy-preview-1346--up-for-grabs-test-bench.netlify.com/

Before I can go live with this, I need to do a bunch of things:
 
 - [x] extract first commit into own PR
 - [x] wire up new action to open a PR which will update these stats
 - [x] find a way to add a tooltip with the last updated field (using a quick "relative time" function)
 - [x] hide projects from results which have 0 cached issues (different case to "unknown")
 - [x] test updating stats again today 
 - [x] move action to run at a different hour each day so it won't clash with project cleanup work

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -30,8 +30,8 @@
     </ul>
   </div>
   <table class="projects">
-   <% _.each(projects, function(project){ %>
-    <tbody>
+   <% _.each(projects, function(project) { %>
+    <tbody class="<% if ( project.stats ) { %>counted<% } %>">
     <tr>
        <td colspan="2" class="title">
            <span class="proj"><a href="<%-project.site %>" target="_blank"><%-project.name %></a></span>
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -39,7 +39,17 @@
     </tr>
     <tr>
        <td class="details">
-        <p class="label"><a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>" target="_blank" tabindex="-1"><%-project.upforgrabs.name %></a></p>
+        <p class="label">
+          <a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>" target="_blank" tabindex="-1">
+            <%-project.upforgrabs.name %>
+
+            <% if (project.stats) { %>
+              <span class="count"><%-project.stats['issue-count']%></span>
+            <% } else { %>
+              <span class="count">??</span>
+            <% } %>
+          </a>
+        </p>
        </td>
        <td class="details">
            <% if (project.desc) { %>
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -42,11 +42,8 @@
         <p class="label">
           <a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>" target="_blank" tabindex="-1">
             <%-project.upforgrabs.name %>
-
             <% if (project.stats) { %>
               <span class="count"><%-project.stats['issue-count']%></span>
-            <% } else { %>
-              <span class="count">??</span>
             <% } %>
           </a>
         </p>
@@ -92,4 +89,3 @@
   pageTracker._trackPageview();
   } catch(err) {}
 </script>
-
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -42,8 +42,12 @@
         <p class="label">
           <a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>" target="_blank" tabindex="-1">
             <%-project.upforgrabs.name %>
-            <% if (project.stats) { %>
-              <span class="count"><%-project.stats['issue-count']%></span>
+            <% if (project.stats) {
+              const now = new Date();
+              let lastUpdated = new Date(project.stats['last-updated']);
+              let since = relativeTime(now, lastUpdated);
+              %>
+              <span class="count" title="Last updated <%- since %>"><%-project.stats['issue-count']%></span>
             <% } %>
           </a>
         </p>
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -20,10 +20,41 @@ define([
       : location.href + 'filters';
   };
 
+  // inspired by https://stackoverflow.com/a/6109105/1363815 until I have a better
+  // idea of what we want to do here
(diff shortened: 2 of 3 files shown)


_includes/scripts.html
```diff
@@ -30,16 +30,27 @@ <h5>Popular tags: </h5>
     </ul>
   </div>
   <table class="projects">
-   <% _.each(projects, function(project){ %>
-    <tbody>
+   <% _.each(projects, function(project) { %>
+    <tbody class="<% if ( project.stats ) { %>counted<% } %>">
```
shiftkey, line 34: We need to set this `class` value to opt-out of querying the GitHub API:

https://github.com/up-for-grabs/up-for-grabs.net/blob/4ded32b6a4af1fd11217f28b22c469810a4d3211/javascripts/main.js#L254-L260

javascripts/main.js
```diff
@@ -20,10 +20,41 @@ define([
       : location.href + 'filters';
   };
 
+  // inspired by https://stackoverflow.com/a/6109105/1363815 until I have a better
+  // idea of what we want to do here
+  function relativeTime(current, previous) {
```
shiftkey, line 25: This function is very rough, and if someone wants to polish this at all they're welcome to

javascripts/projectsService.js
```diff
@@ -218,10 +218,14 @@ define(['underscore'], function(_) {
       }
     }
 
-    var projects = _.map(ordering, function(i) {
+    var all_projects = _.map(ordering, function(i) {
       return _projectsData.projects[i];
     });
 
+    var projects = _.filter(all_projects, function(project) {
+      return project.stats ? project.stats['issue-count'] > 0 : true;
```
shiftkey, line 226: This might be the most controversial change of the PR - hiding projects which have a cached `0` count from the list. I think this saves on users searching for projects that are currently inactive or don't have work,  and I'm open to refining this based on how users feel about this experience.

shiftkey (commented):
Just adding some context to the review not captured by the commits

javascripts/projectsService.js
ritwik12: @shiftkey This could be a good thing in many aspects. But also bad at few aspects such as when all the issues are closed for that repo and the owner or contributors forgets to mark the new issues with the label that they have given at up-for-grabs, then we will lose that Repo. The repo/ project could be good and contributors can reach the project and might open their own issues. It will be good to hide if the repo is not having any issues at all but to consider having no issues based on the label that they provided to us could be wrong. That's one bad aspect out of many could for your decision. Please weigh this with good aspects.

shiftkey: For reference, there are 114 projects listed that currently have no open issues. I'm 👍 to refining this, but I'd like to get this caching work improvements out and we can then see how it feels for users.

> The repo/ project could be good and contributors can reach the project and might open their own issues.

Are they going to be doing this? That's not clear to me based on how users might be arriving at the site, and I worry listing lots of projects without tasks available makes it harder to find ideal projects for contributors.

> It will be good to hide if the repo is not having any issues at all but to consider having no issues based on the label that they provided to us could be wrong.

I've got a good grasp on detecting the incorrect label in https://github.com/up-for-grabs/up-for-grabs.net/pull/1330 which is now bundled into the "updated stats" action with a message like this:

```
Cannot find label 'Help wanted' for project 'matomo-org/matomo'
```

Resolving these requires manually checking the project to see if its due to a spelling mistake, a rename or whether no suitable candidate exists. This could definitely be tweaked based on feedback and potential additional data we need, but it's feeling pretty reliable at the moment.

ritwik12: Sure, we can do that. Thanks for the explanation. Things seem valid. Let's push it @shiftkey

ritwik12 (approved):
We are good to go with this, Thanks for an amazing work here @shiftkey

PR #1353, opened 2019-09-24
Create CXTPL

blockspacer:
C++ template engine that generates C++ code from template, so you will gain VERY good performance and full power of C++.


shiftkey (approved):
@blockspacer thanks!

PR #1365, opened 2019-09-27
Add Ushahidi to projects

rowasc:


_data/projects/ushahidi.yml
```diff
@@ -0,0 +1,13 @@
+name: Ushahidi
+desc: Ushahidi is an open source web application for information collection, visualization and interactive mapping. It helps you to collect info from: SMS, Twitter, Email.
```
ritwik12: @rowasc I guess `:` in your `desc` is causing issues as `:` is used for mapping in .yml

ritwik12 (changes requested):

ritwik12:
@rowasc Please correct the changes that I have suggested to pass all the checks.

_data/projects/ushahidi.yml
rowasc: oops, thank you ! changed

shiftkey (dismissed):
Thanks @rowasc!

_data/projects/ushahidi.yml
```diff
@@ -0,0 +1,13 @@
+name: Ushahidi
+desc: Ushahidi is an open source web application for information collection, visualization and interactive mapping. It helps you to collect info from SMS, Twitter, Email.
+site: https://github.com/ushahidi/platform
+tags:
+- activism
+- non-profit
```
ritwik12, line 6: non-profit and nonprofit means same thing

ritwik12: Can you please also add tags for languages that would be used in this project for contribution? I saw few issues related to code also. Most contributors decide repos on the basis of the skills and coding languages they are aware of., such as JS, Python, Java etc

ritwik12 (changes requested):

_data/projects/ushahidi.yml
ritwik12: Does have two diffferent tags do anything good for your repo? If yes please keep it that way

rowasc: >non-profit and nonprofit means same thing

 I know that... but people use it both ways. I'll just use one since there are no non-profit tags in up-for-grabs apparently at the moment. 

> Can you please also add tags for languages that would be used in this project for contribution

Sure thing, wasn't sure if it was needed or if you pulled it from the repo itself since it wasn't being consistently used in the projects here.

>Does have two different tags do anything good for your repo? If yes please keep it that way

Not sure which two tags.

rowasc: ok. Added more tags for languages, framework, and documentation since I saw it being used in other projects.

ritwik12: I was mentioning `non-profit` and `nonprofit` only by two tags which you already answered.
So, we don't pull tags from repo but from what you define in your .yml project file so that contributors are flexible in creating their own tags.

ritwik12 (approved):

PR #1367, opened 2019-09-27
make this window much shorter to pay down unmergeable PRs

shiftkey:
It's easy enough to resubmit a project after we've left feedback, so I propose shortening the window we leave PRs with feedback open.


ritwik12:
That would be much better and of ease for us

ritwik12 (approved):

PR #1391, opened 2019-10-01
change primary colour for headers and links

tstathara:
change main color

### DIFF
--- a/stylesheets/stylesheet.css
+++ b/stylesheets/stylesheet.css
@@ -68,7 +68,7 @@ h1, h2, h3, h4, h5, h6 {
 h1, h2, h3 {
   font-family: 'Cookie', cursive;  
   margin-bottom: 20px;
-  color: #2E7BA9;
+  color: #225b7d;
 }
 
 h1 {
@@ -301,7 +301,7 @@ footer {
 
 a, .as-link {
   text-decoration: none;
-  color: #2E7BA9;
+  color: #225b7d;
   line-height: 1.4em;
   display: inline-block;
   vertical-align: baseline;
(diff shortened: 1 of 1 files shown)


ritwik12:
@tstathara  Please share a preview for this

shiftkey:
> Please share a preview for this

@ritwik12 every PR that gets opened here has a corresponding Netlify staging site created (if the site builds successfully): 



The `Details` link here will show you what you're looking for.

To illustrate the color change, I went to [this site](https://www.colortools.net/color_compare_colors.html) and put them side-by side:



@tstathara are you able to elaborate on why you chose this colour? I'm really hazy on the context, and the staging site seems nice, but I'd love to understand the context more before merging...

tstathara:
It is the color from the image in header. It is darker, i find it better

shiftkey:
@tstathara looking at the [deploy link](https://deploy-preview-1391--up-for-grabs-test-bench.netlify.com/#/) I can see there's a banner in the top-right that still has the old colour (or similar) for it's background:



That seems to be coming from this value:

https://github.com/up-for-grabs/up-for-grabs.net/blob/f8bad195299233ea224d23ae70ab1ab0171ad06f/stylesheets/stylesheet.css#L150

Are you interested in also updating that style?

tstathara:
of course!

shiftkey:
Having stared at this a bunch over the previous week, I'm still a fan of what we have currently with the two slightly different blues. In particular it gives the banner some depth:



Contrasting with what's proposed in this PR, it feels more one-dimensional and less like a banner:



And the blue we use currently looks more brighter when it's used a lot in the project list:



Switching that over to the darker blue feels a bit heavier:



My design skills are not great, so I don't really have any good advice about how to make this better, but I'm currently feeling like I prefer the current colour scheme over what's proposed by this PR. I wish I felt differently @tstathara!

Going to let this sit until Wednesday to see if anyone else has differing opinions on this.

ritwik12:
@tstathara @shiftkey I prefer the current bright one, don't see a need to switch to dark.

shiftkey:
Closing this out as we prefer the current colour scheme.

Thanks for the contribution @tstathara!

Issue #1368, opened 2019-09-27
Update Readme for New Features


ritwik12:
Since, everyday we are making up-for-grabs.net better and better with new features. It will be good to keep everyone in the same loop so that they will be aware of what all up-for-grabs provides and uses to contribute better.

A good start will be to see the latest merged PRs with new features and Issues solved.

jakkuthompson:
I've edited the README, let me know when I can get assigned.

shiftkey:
> A good start will be to see the latest merged PRs with new features and Issues solved.

I don't agree these are the best things to put into the README, because they're likely to go stale quick. What I'd like to see in the README for this project is the details that cover:

 - an introduction to what the project is about
 - if you're a project maintainer, what's important to know?
 - if you're interested in contribution, what's important to know?
 - if you're looking for more documentation about the project, where is it?

There's some detailed instructions currently in the README that isn't very out of date which I think can be moved out to separate documentation:

 - how to setup your environment and test the site
 - how to scaffold the project
 - how the site gets published
 - what automation is running and what does it do?

I think that would lead to a much easier to follow README. If we're 👍 on this direction I can open some issues to track the specific tasks I had in mind,  so others can help out.

ritwik12:
@shiftkey I think I made it confusing, By `A good start will be to see the latest merged PRs with new features and Issues solved.` I meant the code changes that we do to make up-for-grabs better like the latest `stats` feature. I never meant adding details about projects being added to up-for-grabs. That's irrelevant.

I agree with you let's open issues.

shiftkey:
@ritwik12 with #1614 merged I'm feeling pretty good about how the current `README` summarises and introduces the project. If we have more specific ideas for the README, I think these should be separate and more-specific discussions, like #1393.

Are we good to close this out?

ritwik12:
@shiftkey Lgtm!!

PR #1394, opened 2019-10-01
tidy up README, add section on automation

shiftkey:
Related to #1368 

I wanted to add some context about what we're automating and why it's important. I also tidied up some other bits while I was in there.


shiftkey:
Gonna merge this in as it's now a blocker on #1413

ritwik12:
@shiftkey Thanks for merging, looks cool :)

PR #1403, opened 2019-10-01
Added lang="en" to <html>

davidhaw:
When I was running the Google Chrome Audit on up-for-grabs.net, I saw that one of the errors given was that it did not include a lang attribute in <html>. I quickly fixed that by adding lang="en" in the main layout.


shiftkey (approved):
@davidhaw thanks!

PR #1404, opened 2019-10-02
added meta name="robots" / update "chosen css" version link

slartibartfass:


_includes/head.html
```diff
@@ -33,7 +34,7 @@
   <meta name="msapplication-TileColor" content="#2b5797">
   <meta name="msapplication-TileImage" content="{{ site.github.url }}/icons/mstile-144x144.png">
 
-  <link rel="stylesheet" type="text/css" media="screen" href="https://cdnjs.cloudflare.com/ajax/libs/chosen/1.0/chosen.min.css">
+  <link rel="stylesheet" type="text/css" media="screen" href="https://cdnjs.cloudflare.com/ajax/libs/chosen/1.8.7/chosen.min.css">
```
shiftkey, line 37: This is a great spot - thanks for noticing. For reference, we're using 1.8.7 of the chosen JS file here:

https://github.com/up-for-grabs/up-for-grabs.net/blob/afe931ec15f043658aa4923c88b2f3a0cba1b8ad/javascripts/app.js#L11

I tested out the deploy URL and the selects look to be styled correctly, but we can easily revert this PR if we spot something weird.

shiftkey (approved):
Thanks @slartibartfass!

PR #1079, opened 2018-10-31
[WIP] render list of projects server-side to support clients who disable JavaScript

mrtnmgs:
Fixes #511

Hi,
Not sure why the diff is highlighting the last line (44)... Does it matter?

PR #1412, opened 2019-10-02
render some projects when browser does not support JS

### LINKED ISSUE #511: Visiting the site with javascript disabled does not display any projects

shiftkey:
Resolves #511
Supersedes #1079

This PR improves the experience when a user visits the site without JS. This is what you currently see today:

I wanted to pick up #1079 and take it to it's logical conclusion but I had some issues with the approach chosen:

 - rendering every project in the list wasn't optimal now that we're at almost at 800 projects
 - the order of projects is not randomized in that way
 - we lacked any sorts of stats for the projects listed (now available due to #1346)

So what does this PR give us now when JS is disabled? 

 - render a randomized list of 50 projects 
   - this is set at build time, so it'll regenerate each time the site publishes to GitHub Pages
 - embedded stats alongside each project 
   - we couldn't do this before because we hit the GitHub API directly)
 - added a warning message to explain the behaviour
   - this is unstyled for now and would be great for someone else to polish

Here's a quick screenshot:

This feels like a better incremental change than what was initially proposed in #1079, and we can refine this over time based on how we feel about it.

 - [x] update comment to link to this PR
 - [x] measure overhead of including this static list of projects
 - [x] confirm or give up on excluding projects with zero open issues
 - [x] test deploy site with JS disabled
 - [x] test deploy site with JS enabled
 - [x] get tag links working or disable them for now

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,8 +19,8 @@ github: [metadata]
 
 # DO NOT DELETE THIS
 # This is used to build up the array of projects rendered inside a <noscript>
-# tag to support browsers which disable JS. For more context check out 
-# https://github.com/up-for-grabs/up-for-grabs.net/pull/XYZ which explains what
+# tag to support browsers which disable JS. For more context check out
+# https://github.com/up-for-grabs/up-for-grabs.net/pull/1412 which explains what
 # this gives us.
 emptyArray: []
 
--- a/_config.yml
+++ b/_config.yml
@@ -17,3 +17,10 @@ exclude:
 
 github: [metadata]
 
+# DO NOT DELETE THIS
+# This is used to build up the array of projects rendered inside a <noscript>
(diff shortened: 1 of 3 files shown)


shiftkey:
With this PR, we only increase the download size of the main HTML file (gzip compressed) by 5KB

Current site:



This PR:



Over in https://github.com/up-for-grabs/up-for-grabs.net/pull/1079#issuecomment-462453073 we were looking at an additional 70KB being served up to every visitor, so this feels like a great improvement.

PR #1413, opened 2019-10-02
update to bundler 2

shiftkey:
I'm having to carry around this change to work outside of Docker, so I'd like to see what else breaks by moving us up to [Bundler 2](https://bundler.io/blog/2019/01/03/announcing-bundler-2.html) which has been out since January and mentions supporting "version autoswitching", whatever that means...

 - [x] bump lockfile
 - [x] bump netlify environment
 - [x] figure out why Netlify doesn't pull in the `BUNDLED_WITH` version - https://github.com/netlify/build-image/issues/296
 - [x] update docs


shiftkey:
To wrap this up I need to:

 - dive into netlify config and choose this image (can't control this via config)



 - re-run the build so the deploy passes
 - merge it
 - handle any older PRs that might break due to the older bundler version

PR #1424, opened 2019-10-04
Create_containers.yml

ping2varshit:


_data/projects/containers.yml
```diff
@@ -0,0 +1,10 @@
+name: containerpilot
+desc: 'An init system for cloud-native distributed applications'
+site: https://github.com/joyent/containerpilot
+tags:
+- containers
+- init
+- orchestration
+upforgrabs:
+  name: help needed 
+  link: https://github.com/ping2varshit/containerpilot
```
shiftkey, line 10: It looks like this is linking to your fork of the project, which doesn't have any open issues.

If you'd like to see this project listed on the site, we need:

 - a label that's applied to issues that are ideal for new contributors (previous line)
 - the URL to this list of issues so the user can easily find these issues (this line)

Without those two things, I cannot merge this PR.

shiftkey (changes requested):
To ensure this doesn't get merged early

PR #1435, opened 2019-10-05
remove inline JSON and globals for projects

shiftkey:
Following on from #1434, this PR removes the inline projects list and introduces a standalone module to asynchronously load and setup the list of projects.

 - [x] investigate how this changes download sizes
 - [x] investigate how this changes performance of site

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -69,17 +69,6 @@
    <% }) %>
    </table>
 </script>
-<script src="//cdnjs.cloudflare.com/ajax/libs/showdown/1.3.0/showdown.min.js"></script>
-<script type="text/javascript">
-  var files = {{ site.data.projects | jsonify }};
-  var projects = new Array();
-  var converter = new showdown.Converter();
-  for (var fileName in files) {
-    var project = files[fileName];
-    project.desc = converter.makeHtml(project.desc);
-    projects.push(project);
-  }
-</script>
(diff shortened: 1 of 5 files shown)


shiftkey:
First set of metrics from testing in Chrome

The size of `index.html` is currently over 300KB uncompressed:



Now it's down to 9KB uncompressed because we've removed the JSON payload and blocking script:



The summary of overall network traffic in production when loading the site (caching disabled):

 - 29 requests
 - 249 KB transferred
 - 723 KB resources
 - Finish: 712 ms
 - DOMContentLoaded: 257 ms
 - Load: 910 ms

With this PR:

 - 31 requests
 - 246 KB transferred
 - 724 KB resources
 - Finish: 868 ms
 - DOMContentLoaded: 177 ms
 - Load: 404 ms

So while `Finish` is now a bit later (because we now asynchronously load the projects over the network), the `Load` and `DOMContentLoaded` timings are lower (because we no longer embed the projects or do the setup inside `index.html`).

shiftkey:
Using the Lighthouse tools in Chrome to measure performance, these were the key stats for Desktop (simulated 4G):



With this PR, we get a good bump in the measurements at the start of the lifetime of the page (left . column):



Testing for Mobile was also similar:

Screen Shot 2019-10-05 at 3 58 24 PM
Screen Shot 2019-10-05 at 3 58 01 PM

PR #1437, opened 2019-10-05
Rubocop errything

divymakkar:
v


shiftkey:
This is not ready for general use

PR #1440, opened 2019-10-06
Use faster ECMAScript 6 `includes` and a simple string concatenation.

CGS-Jack-Bashford:
ES6 is already in use in this file, so this will make the most of it, and string concatenating the ternary is more concise than having a ternary containing string concatenation.

### DIFF
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -15,7 +15,7 @@ define([
 
   var getFilterUrl = function() {
     return (
-      location.href + (location.href.includes("/#/filters") ? "" : "filters")
+      location.href + (location.href.includes('/#/filters') ? '' : 'filters')
     );
   };
 
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -14,7 +14,7 @@ define([
     projectsPanel = null;
 
   var getFilterUrl = function() {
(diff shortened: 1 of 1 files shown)


javascripts/main.js
```diff
@@ -14,9 +14,7 @@ define([
     projectsPanel = null;
 
   var getFilterUrl = function() {
-    return location.href.indexOf('/#/filters') > -1
-      ? location.href
-      : location.href + 'filters';
+    return location.href + (location.href.includes('/#/filters') ? "" + "filters");
```
shiftkey: There's something not right here which is why CI isn't currently passing:

```
$ npm run lint

/home/runner/work/up-for-grabs.net/up-for-grabs.net/javascripts/main.js
  17:82  error  Parsing error: Unexpected token )
```

What were we trying to do here?

CGS-Jack-Bashford: Hi @shiftkey,

I've fixed the typo in my code - see the latest commit.

Jack

shiftkey:
@CGS-Jack-Bashford looks like my review and approval of #1486 means this PR is less valuable now. I'm going to close this out but thanks again for taking a shot at it.

If you're interested in contributing other JS improvements I'll see if I can open some issues later this week.

PR #1472, opened 2019-10-08
Update README.md

snehasis139:


shiftkey:
@kazukilapislazuli thanks for contributing this change. Unfortunately I think we need to think about and tidy up what's currently in here, rather than add more words. 

Please follow along with the discussion in #1368.

PR #1486, opened 2019-10-09
Refactor to template literals

TeslaAdis:


javascripts/main.js
```diff
@@ -16,7 +16,7 @@ define([
   var getFilterUrl = function() {
     return location.href.indexOf('/#/filters') > -1
       ? location.href
-      : location.href + 'filters';
+      : `${location.href}${filters}`;
```
shiftkey: I think there's a subtle bug here because we dropped the quote marks around `filters`, turning it into a local variable and causing this error when you click on any tags:



I think this is what it should be, but I'm not 100% sure:

```suggestion
      : `${location.href}filters`;
```

shiftkey (changes requested):

javascripts/main.js
TeslaAdis: Not sure how I missed this... thanks

shiftkey (approved):
Thanks @TeslaAdis!

shiftkey:
@all-contributors please add @TeslaAdis for code

PR #1494, opened 2019-10-09
Refactor and simplify Project loader

TeslaAdis:
Simplify code, use latest es features

### DIFF
--- a/javascripts/projectLoader.js
+++ b/javascripts/projectLoader.js
@@ -1,4 +1,5 @@
 /* eslint prefer-arrow-callback: [ "error" ] */
+/* eslint arrow-parens: [ "error", "as-needed" ] */
 /* eslint global-require: "off" */
 /* eslint block-scoped-var: "off" */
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@TeslaAdis thanks for opening this PR. Unfortunately there's a bunch of checks failures in the because we've been using eslint and I think this rule is to blame:

https://github.com/up-for-grabs/up-for-grabs.net/blob/3e41c42fa85b11f0b2db1d4b22a39e7427b00457/.eslintrc.json#L12

If we want to take a shot at migrating files one-at-a-time, you could add in the new value at the head of this file and see if it improves things:

```js
```

shiftkey:
@TeslaAdis looks like this is close, but Prettier's formatting isn't correct.

If you run this locally and commit the changes, it should pass CI: `npm run prettify`

shiftkey:
@TeslaAdis I pushed one more commit to this branch to address the area where I believe you got stuck - eslint and prettier disagreeing with eachother about whether to wrap the single-parameter function in parentheses.

I'll open some more issues over the weekend to pay down our eslint rule customizations so we can bring the JS more up-to-date. 

Thanks again for this PR!

shiftkey (approved):

PR #1505, opened 2019-10-10
Fix document changes

varadaprasanth:
Added one new test config file in my branch (fix_document_changes branch) and merging it.


shiftkey:
@varadaprasanth it's not clear to me what problem this is addressing. Can you elaborate?

PR #1519, opened 2019-10-11
chore(javascripts): prefer-arrow-callback

firfircelik:
according to this issue https://github.com/up-for-grabs/up-for-grabs.net/issues/1515.

npm fix everything, sorry for that :)

### DIFF
--- a/javascripts/projectsService.js
+++ b/javascripts/projectsService.js
@@ -45,7 +45,9 @@ define(['underscore'], _ => {
       )
     );
 
-    return _.filter(projects, project => _.contains(projectNames, project.name));
+    return _.filter(projects, project =>
+      _.contains(projectNames, project.name)
+    );
   };
 
   /*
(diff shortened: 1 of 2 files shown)


shiftkey:
> npm fix everything, sorry for that :)

Can you please follow the instructions in the original issue? There are a lot of changes introduced that are separate to the problem outlined, and there are CI issues introduced too...

javascripts/projectsService.js
```diff
@@ -1,5 +1,6 @@
 /* eslint global-require: "off" */
 /* eslint block-scoped-var: "off" */
+/* eslint prefer-arrow-callback: [ "error" ] */
```
shiftkey: Like we've found in #1520, I think we also need to add this, run `npm run prettier-fix` and then commit to make CI pass:


```suggestion
/* eslint prefer-arrow-callback: [ "error" ] */
/* eslint arrow-parens: [ "error", "as-needed" ] */
```

firfircelik: this is the running one. don't get why its failed.

shiftkey:
I think the errors you're now seeing (which I didn't expect) are due to clashes between eslint and prettier.

Add these two rules to the top of the file, and I'll write up an issue to track how we can avoid these clashes in the future:

```js
/* eslint function-paren-newline: [ "off" ] */
/* eslint implicit-arrow-linebreak: [ "off" ] */
```

shiftkey:
@firfircelik thanks for working through these issues with me.

I'm gonna pause and revisit how we use eslint and prettier to see if we can reduce the pain here with upgrading to prevent these "feedback loop" events from happening again.

PR #1520, opened 2019-10-11
chore(javascripts): add prefer-arrow-callback to main.js

firfircelik:
according to https://github.com/up-for-grabs/up-for-grabs.net/issues/1516

### DIFF
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -1,4 +1,5 @@
 // @ts-nocheck
+/* eslint prefer-arrow-callback: [ "error" ] */
 
 define([
   'jquery',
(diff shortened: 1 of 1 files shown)


javascripts/main.js
```diff
@@ -1,4 +1,5 @@
 // @ts-nocheck
+/* eslint prefer-arrow-callback: [ "error" ] */
```
shiftkey, line 2: I think if you pull in this change, run `npm run prettier-fix` and then commit this should make CI pass:

```suggestion
/* eslint prefer-arrow-callback: [ "error" ] */
/* eslint arrow-parens: [ "error", "as-needed" ] */
```

PR #1526, opened 2019-10-11
Correct Grammar

SteveLee1287:
Replace does with do

### DIFF
--- a/docs/client-side-scripting.md
+++ b/docs/client-side-scripting.md
@@ -5,7 +5,7 @@ found and run in the app.
 
 Why do this?
 
-- GitHub Pages does not support JS bundles natively, and will block any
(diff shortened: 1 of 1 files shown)


docs/client-side-scripting.md
```diff
@@ -5,7 +5,7 @@ found and run in the app.
 
 Why do this?
 
-- GitHub Pages does not support JS bundles natively, and will block any
+- GitHub Pages do not support JS bundles natively, and will block any
```
shiftkey, line 8: I believe this is correct as it refers to GitHub Pages (the product) rather than a plural number of pages

SteveLee1287: That makes sense. My apologies I read it in the context of actual number of pages!

PR #1530, opened 2019-10-11
Add up-for-grabs configuration for exiv2

D4N:


_data/projects/exiv2.yml
```diff
@@ -0,0 +1,19 @@
...
+- c++
+- c++11
+- image
+- metadata
+- exif
+- iptc
+- xmp
+- cross-platform
+upforgrabs:
+  name: "good first issue"
```
shiftkey: No need to wrap this in quotes unless there are some special characters

```suggestion
  name: good first issue
```

shiftkey (changes requested):
@D4N one suggestion to tidy this up, otherwise it looks good to me!

_data/projects/exiv2.yml
D4N: Fixed!

shiftkey (approved):

shiftkey (approved):
@D4N thanks!

PR #1539, opened 2019-10-13
scaffolding for a 'unified' actions environment

shiftkey:
This is a little experiment to poke at the Actions environment to see if I can provide it a named `Dockerfile` in a directory (up until this point I've only provided the directory name, so this might not work).

Why does this matter? When `docker build` runs it uses the directory containing the Dockerfile as the context, and scripts outside the directory are inaccessible (so I can't copy in files I might need). This is great for simple things, but I'm building up some common Ruby functionality that I'd like to share between actions, without going through the ceremony of publishing to rubygems.

### DIFF
--- a/.github/actions/all/sitevalidator.Dockerfile
+++ b/.github/actions/all/sitevalidator.Dockerfile
@@ -1,0 +1,7 @@
+FROM ruby:2.6-stretch
+
+RUN pwd
+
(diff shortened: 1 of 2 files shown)


shiftkey:
Screen Shot 2019-10-13 at 11 38 56 AM

PR #1543, opened 2019-10-14
Create ceaser_shift.yml

sainad2222:
This project is about encryption and decryption of caesar shift

### DIFF
--- a/_data/projects/ceaser_shift.yml
+++ b/_data/projects/ceaser_shift.yml
@@ -1,0 +1,15 @@
+name: ceasar_shift
+desc: It is used to encrypt and decrypt text in caesar cipher
+site: https://github.com/sainad2222/ceaser_crypto
+tags:
+-crypto
+-caesar shift
+-cipher
+-encryption
+-decryption
+upforgrabs:
+  name: help needed
+  link: https://github.com/sainad2222/ceaser_crypto/blob/master/ceaser.py
+stats:
+  issue count=0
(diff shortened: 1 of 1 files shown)


_data/projects/ceaser_shift.yml
```diff
@@ -0,0 +1,15 @@
...
+site: https://github.com/sainad2222/ceaser_crypto
+tags:
+-crypto
+-caesar shift
+-cipher
+-encryption
+-decryption
+upforgrabs:
+  name: help needed
+  link: https://github.com/sainad2222/ceaser_crypto/blob/master/ceaser.py
```
shiftkey, line 12: This link should be to a curated list of issues that contributors can help out with - is this something you're able to put together?

sainad2222: Need improvements in code like efficiency and speed

shiftkey: > Need improvements in code like efficiency and speed

Please create issues that outline these areas of improvement, the things that contributors should know about, and how they can help.

From the footer of the Up-for-Grabs site:

> We're looking for projects who can take the time out to help mentor developers as they get started with open source.
> 
> What sort of tasks are a good fit?
>
> - Tasks should take no longer than a few nights' worth of work
> - Tasks should stand alone - avoid core functionality on which other tasks depend
> - Tasks should be well described with pointers to help the implementer

If that doesn't sound like a good fit, maybe it's not a good listing for this site

PR #1577, opened 2019-10-17
grammar issue fixed

mahditakrim:
A grammar issue in before.html line 5 (first item of bulleted list) fixed by replacing the word 'for' with 'of' .

### DIFF
--- a/_includes/before.html
+++ b/_includes/before.html
@@ -11,7 +11,7 @@
  Find a project you'd like to get involved with:
 
   <ul>
-    <li>Read the contributor guidelines for the project</li>
+    <li>Read the contributor guidelines of the project</li>
     <li>Get the project running locally</li>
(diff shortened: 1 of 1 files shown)


shiftkey:
@mahditakrim thanks for opening this PR!

I was a bit indecisive about this change so I went hunting for a good resource on prepositions and how to best measure this, and I ended up on https://www.talkenglish.com/grammar/prepositions-of-to-for.aspx.

This seems to indicate `of` is indeed the better term here. From that post. both of these cases look relevant for us:

> **Of**
> Used for belonging to, relating to, or connected with:
> ...
> Used to indicate reference:
> ...

These feel more suitable than any of the recommended cases for `for`:

> **For**
> Used to indicate the use of something:
> ...
> Used to mean because of:
> ...
> Used to indicate time or duration:

shiftkey (approved):

PR #1585, opened 2019-10-19
update

Ricehardt:


shiftkey:
@Aeolyx this list gets updated by our @all-contributors integration.

See #1571 for an example.

PR #1589, opened 2019-10-19
add names to config jobs

shiftkey:
When I get a notification that a build fails I get GUIDs for the job names:

I think this is the way to make the list show a more friendly name.


shiftkey:
No change, oh well. I'll pass that feedback onto the team when I get a chance.

Screen Shot 2019-10-19 at 12 02 20 PM

PR #1593, opened 2019-10-20
ADDITION FOR DATA SCIENCE ANALYSTS

snehit31:
Hi! This is file was added to get the linking of all data analysts interested to create a community to help each other and collaborate constructively to form something significant.


shiftkey:
```
1 files found in root which look like content files:
 - Data_Science.yml
Move these inside _data/projects to ensure they are listed on the site
```

PR #1594, opened 2019-10-21
PLACE FOR ALL DATA SCIENCE ENTHUSIASTS.yml

snehit31:
I would like to call upon all data science enthusiasts present in the community to use it like a platform to create something great.

PR #1595, opened 2019-10-21
DATASCIENCEENTHUSIAST.yml

snehit31:
I would like to call upon all data science enthusiasts present in the community to use it like a platform to create something great.

### DIFF
--- a/_data/_DATASCIENCEENTHUSIAST.yml
+++ b/_data/_DATASCIENCEENTHUSIAST.yml
@@ -1,0 +1,12 @@
+
+name: Data Science
+desc: It is for data science enthusiasts like me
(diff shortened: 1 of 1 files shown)


shiftkey:
This seems to be the same project as #1594 and #1593, but the YAML file is not readable by Jekyll:

```
jekyll 3.8.5 | Error:  (/home/runner/work/up-for-grabs.net/up-for-grabs.net/_data/_DATASCIENCEENTHUSIAST.yml): could not find expected ':' while scanning a simple key at line 10 column 1
```

shiftkey:
There are different errors between these three PRs #1593 #1594 #1595, and for the sake of simplicity I'm going to close these out in favour of the first PR #1593.

PR #1605, opened 2019-10-22
Update index.html

DHStanLingRay21:
changed "panel" to "panels"


shiftkey:
@DHStanLingRay21 this change will break JavaScript that looks for this ID

PR #1616, opened 2019-10-24
Add faas.yml

JoaoManoel:


_data/projects/faas.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: faas
```
shiftkey: This name looks very general. Should it be more specific?

```suggestion
name: OpenFaaS
```

_data/projects/faas.yml
```diff
@@ -0,0 +1,15 @@
...
+desc: OpenFaaS - Serverless Functions Made Simple
```
shiftkey: If you prefer the name above, this can be simplified:

```suggestion
desc: Serverless Functions Made Simple
```

shiftkey (commented):
@JoaoManoel thanks for submitting!

Changes look good, just a couple of suggestions about the metadata.

shiftkey (approved):
@JoaoManoel thanks!

PR #1626, opened 2019-10-25
Add Kwik to project list

jcornaz:
Kwik is a property-based testing library which I think could both benefit from and be beneficial to open-source contributors.

I have a bunch of small issues that can be done by anyone having enough interest in Kotlin and property based testing.

I think Even more of them will arise over time, since I will often have something like "create this new fancy random generator", which will always be easy and a good start for new contributors.

I hope you'll like it, and I'd be glad to make it easier to contribute, if you have advises to give.

### DIFF
--- a/_data/projects/kwik.yml
+++ b/_data/projects/kwik.yml
@@ -1,0 +1,11 @@
+name: Kwik
+desc: A property-based testing library for Kotlin
+site: https://kwik.readthedocs.io
+tags:
+  - kotlin
+  - library
+  - test
+  - testing
+upforgrabs:
+  name: up for grabs
+  link: https://github.com/jcornaz/kwik/issues?q=is%3Aissue+is%3Aopen+label%3A%22up+for+grabs%22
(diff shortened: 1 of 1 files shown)


_data/projects/kwik.yml
```diff
@@ -0,0 +1,10 @@
+name: Kwik
+desc: A property-based testing library for Kotlin
+site: https://kwik.readthedocs.io
+tags:
+  - kotlin
+  - libraries
+  - testing
+upforgrabs:
+  name: up for grabs
+  link: https://github.com/jcornaz/kwik/issues?q=is%3Aissue+is%3Aopen+label%3A%22up+for+grabs%22
```
shiftkey: There's a nicer link that can be used here:

```suggestion
  link: https://github.com/jcornaz/kwik/labels/up%20for%20grabs
```

shiftkey (changes requested):
@jcornaz thanks for submitting! Just one change requested, everything else looks great!

_data/projects/kwik.yml
jcornaz: Indeed. Thanks.

shiftkey (approved):

PR #1637, opened 2019-10-28
Update README.md

k1sh0rek0ld:
Starting with Caps


shiftkey:
@k1sh0rek0ld I'm not sure this really adds much to the page. Thoughts?

ritwik12:
@k1sh0rek0ld I guess the original is good and we would like better contributions from you in future :)

ritwik12:
@k1sh0rek0ld Thanks !!

Issue #930, opened 2018-05-07
Improving Frontend


ritwik12:
At present, up-for-grabs.net is great content wise and is also good in front-end, But I can see a lot of potential for improvement as there are so many great web designers who can make up-for-grabs.net a very beautiful website. I would recommend [awwwards](https://www.awwwards.com/) for getting ideas.

Few points to work on as per @camilosampedro suggestions in the comments:

- [ ] Add task filter by last update (First of all active).
- [x] Сhange layout of the page, in particular, move filters to the side menu. Thereby solving the problem of unused space.
- [x] Add dark (night) mode.
- [ ] Fix dark mode contrast.
- [ ] Set nicer colors and fonts.

NjalTheKnut:
I have an idea that might help tremendously.  I could attempt to implement a scrolling "back to top" button that would allow users who have scrolled way down the page to get back up to the search bar without needing to scroll all the way back manually.  Other suggestions may come to me as I use the site.

ritwik12:
@NjalTheKnut That will be a good feature, go for it.

sunilkumarvalmiki:
hi,can i undertake the work proposed by NjalTheKnut...

NjalTheKnut:
I actually found out how to do it recently.  I am willing to do it and have a second person go over the code to ensure I am doing things in the best fashion possible.

Being somewhat inexperienced with remote team mechanics, I would appreciate a little guidance acquiring the code and making certain I don't foul up production branches in the VCS.

Also, am I allowed to introduce jQuery, if it is not already used in this project?

sunilkumarvalmiki:
[up-for-grabs.net-gh-pages 2.zip](https://github.com/up-for-grabs/up-for-grabs.net/files/2246438/up-for-grabs.net-gh-pages.2.zip)
-------------------------------------------------------------

![img](https://user-images.githubusercontent.com/24809771/43476735-ed64f6a8-9516-11e8-8520-b0f60ba1e00b.png)
------------------
i have completed the task...

ritwik12:
@sunilkumarvalmiki  Can you host this on your fork? it will be easy to review that way.

sunilkumarvalmiki:
yes, i forked it to my profile.

ritwik12:
@NjalTheKnut I meant if you can host this as github pages on your fork so that we can see it live.

Supreme-Kai:
[...]

PR #1642, opened 2019-10-28
Improve hover on button and h1 readablility

sajjadcodes:
I have just improved the readability of heading as well as a made href more interactive.

### DIFF
--- a/stylesheets/stylesheet.css
+++ b/stylesheets/stylesheet.css
@@ -72,7 +72,7 @@ h1, h2, h3 {
 }
 
 h1 {
-  font-size: 56px;
+  font-size: 58px;
 }
 
 .block h1 {
@@ -564,7 +564,11 @@ form {
   display: inline-block;
   padding: 2px 8px;
 }
-
+.projects .label a:hover {
+  color: #FFF;
+  font-size:1.1em;
+  padding: 4px 10px;
+}
 .projects .tags {
   font-size: 12px;
   font-style: italic;
--- a/stylesheets/stylesheet.css
+++ b/stylesheets/stylesheet.css
@@ -72,7 +72,7 @@ h1, h2, h3 {
 }
 
 h1 {
-  font-size: 58px;
+  font-size: 56px;
 }
 
 .block h1 {


sajjadcodes:
Improvement in readibility

stylesheets/stylesheet.css
```diff
@@ -72,7 +72,7 @@ h1, h2, h3 {
 }
 
 h1 {
-  font-size: 56px;
+  font-size: 58px;
```
ritwik12: @sajadhussain Please remove this as it is not having any valid changes.

ritwik12:
@sajadhussain Would like to keep the h1 as original only and please remove that change from this PR. I would like to keep hover in this PR as it looks good. @shiftkey Hover looks good but you can decide the exact px and size of hover.

ritwik12 (changes requested):

stylesheets/stylesheet.css
sajjadcodes: okay got

shiftkey:
> @shiftkey Hover looks good but you can decide the exact px and size of hover.

@ritwik12 @sajadhussain I'm curious what problem we're trying to address here with this change. I can see what's changed but I'd like to understand _why_ we're doing it this way, as we might be able to provide useful feedback about the changes or suggest a better solution.

ritwik12:
@sajadhussain Please let us know what problems it will solve and why is it needed. Better to have an issue for this to discuss further.

sajjadcodes:
It is UX improvement of the site. Just improve the button. code hover so it will more interactive and give a sense of working.

ritwik12:
@sajadhussain Thanks for your PR, it might seem good but is not that much needed here as the primitive site looks good in itself.

We have many issues that need attention and would be happy if you can help there based on your skills. like #930, etc

@shiftkey Please feel free to give your thoughts here

shiftkey:
> Please feel free to give your thoughts here

There's a few things I don't like about this PR, and I've tried to capture this in a GIF so you can see it for yourself:

 - when you hover over the button, the text alongside it gets nudged down slightly because of the padding changes. 



 - making the font size slightly bigger at the same time feels a bit jarring

 - this solution leverages `:hover`, which is only available for desktop users - what about mobile users, or users with screen readers?

> It is UX improvement of the site. Just improve the button. code hover so it will more interactive and give a sense of working.

In situations like this I'd prefer to discuss the problem through an issue, rather than going straight to a proposed solution using a pull request. The underlying problem is still unclear to me, and I believe using animations like this for web pages should only occur when absolutely necessary.

If we've got a good grasp on the problem and nothing else seems suitable, then I'm fine to revisit this.

PR #1643, opened 2019-10-28
adding conrad

Sangarshanan:
Conrad is a simple CLI tool to track conferences and meetups on your terminal 

I think it would be a great addition to up-for-grabs : )


shiftkey (approved):
@Sangarshanan thanks!

PR #1648, opened 2019-10-29
Removing peoples' VSCode configs and ignoring them

dabockster:
Someone left their personal VSCode config in here, so I removed it and added .vscode/ to the gitignore.


shiftkey:
@dabockster thanks for the contribution, but the `launch.json` file contains scripts to enable you to run and debug the JS tests locally.



This was very much intended.

shiftkey:
@dabockster I'm going to close this PR because it was intentional to include that file, but happy to chat more about it (either here or in a fresh issue).

PR #1649, opened 2019-10-29
 Added bundler to the gem install script in Dockerfile 

dabockster:
I know Jekyll has bundler as a dependency already. But I've had issues on several Windows installs where the Jekyll package script won't build bundler correctly. I think it has something to do with native libraries being used. However, it builds just fine on its own. So I put it in the gem installer command before it calls Jekyll package to mitigate this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ENV JSON_GEM_VERSION 1.8.6
 RUN apk --update add --virtual build_deps \
     build-base ruby-dev libc-dev linux-headers \
   && gem install --verbose --no-document \
+    bundler \
     json:${JSON_GEM_VERSION} \
     github-pages:${GITHUB_GEM_VERSION} \
     jekyll-github-metadata \


shiftkey:
> But I've had issues on several Windows installs where the Jekyll package script won't build bundler correctly. 

Can you share some more context about how this is failing? I'm happy to help troubleshoot further.

I'm a bit puzzled why the docker image, which is based on Alpine Linux, would care about the underlying OS. But maybe this is some corner of Docker for Windows that I'm not familiar with.

Also, shouldn't this be installing the same version that's expected in the `Gemfile.lock`? This currently is installing the latest version (2.0.2), and while I think Bundler v2 is more resilient to different minor versions it'd be great to avoid users accidentally committing version bumps if possible:

https://github.com/up-for-grabs/up-for-grabs.net/blob/287d513e5a196523b2913d83a6458c8296f6d5f7/Gemfile.lock#L300-L301

shiftkey:
Closing this out due to inactivity. @dabockster if you were still interested in landing this fix, please open a fresh PR and address the questions above.

PR #1656, opened 2019-10-31
#ffffff is not necassary. #fff can be used

Denuka1993:


shiftkey:
@Denuka1993 thanks for the contribution!

> #ffffff is not necassary. #fff can be used

While this is definitely valid, I prefer to be consistent with our CSS colours and use the same length for these hex codes.

PR #1657, opened 2019-10-31
Update mentions of project name

Rinsington-DC:
Changed from for to For to be consistent with brand

 - [ ] update mentions in `README.md`
 - [ ] update mentions in `docs/list-a-project.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ If you would like to get involved with the project, please read the
 
 ## How does the site work?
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@Rinsington-DC there's also a change on line 13 here that we should address while we're in here:

https://github.com/up-for-grabs/up-for-grabs.net/blob/b45d23d31fd3457f7534ba57b9ce0513a94da9f5/README.md#L13

PR #1665, opened 2019-11-03
[WIP] update and merge stats change in background, without opening a PR

shiftkey:
It's been over a month since we first enabled stats in the project, and I think I'm ready to leave this running in the background rather than have it open a PR after each run.

- [x] test changes locally before doing any further work
- [x] commit and push to branch matching pattern
- [x] stop creating PR after work is done
- [x] update CI action to run on branches matching this pattern
- [x] create action to run on [`status`](https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows#status-event-status) event
  - [ ] exit if not a `success` event -> something went wrong
  - [ ] find PR associated with branch -> exit if not found
  - [ ] if PR is mergeable, use the [merge API](https://developer.github.com/v3/repos/merging/) to land the commits -> exit if not valid
- [ ] ???

### DIFF
--- a/.github/actions/merge-successful-branch/Dockerfile
+++ b/.github/actions/merge-successful-branch/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:2.6-stretch
+
+# Labels for GitHub to read the action
+LABEL "com.github.actions.name"="Merge successful branch"
+LABEL "com.github.actions.description"="When a commit status returns success, merge the related branch if it belongs to one of our scheduled actions."
+LABEL "com.github.actions.icon"="git-merge"
+LABEL "com.github.actions.color"="purple"
+
+COPY Gemfile ./
+
+RUN bundle version
+
+RUN bundle install
+
+COPY . .
+
+ENTRYPOINT ["/run.sh"]
--- a/.github/actions/merge-successful-branch/Gemfile
+++ b/.github/actions/merge-successful-branch/Gemfile
@@ -1,0 +1,6 @@
+source 'https://rubygems.org'
+
+gem 'octokit'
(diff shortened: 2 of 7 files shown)


ritwik12:
@shiftkey One doubt, Do we see a PR merged after this or there will be no PR record as such?
And if we don't have a PR record then why simply not a cron job for this task? I know having a Cron job on a public repo is not a good idea but still just a thought.

shiftkey:
> One doubt, Do we see a PR merged after this or there will be no PR record as such?

My current thinking is that I don't see much value in the PR being opened and merged beyond generating notifications - if we we're not reviewing it much beyond "CI passes" and we have confidence in the automation it feels like noise we can live without.

> And if we don't have a PR record then why simply not a cron job for this task?

The job here can only do so much currently:

 - actions are run when a PR is opened (or on the default branch), but we can enable it for all branches
 - the Netlify deploy preview would be great to run in this situation, and I think I can enable that but I haven't confirmed that's the case
 - the hook to do work when both of these succeed would be the `status` event

I've put this on hold because of some limitations I've uncovered that might impact what we can do here:

 - actions being run from PRs for forks do not have write access to the API (for security reasons), and I'm not sure how the `status` event would work for a commit from a fork
 - I also wanted to preserve the squash-and-merge options here, and that requires opening a pull request to access that feature
 - if I eventually want to enable branch protections on `gh-pages` again to ensure everything goes through PRs the `github-actions` account does not have the ability to merge the PR (I can't add it as a collaborator)

shiftkey:
Closing this out for now because I plan to go in a different direction with this

PR #1690, opened 2019-11-14
Introduce Finance Manager

josharsh:


shiftkey (approved):
@josharsh thanks for the contribution!

I wanted to test this action with some live PRs but it looks like there's something unexpected happening with the "action being run from a fork" that I wasn't ready for.

Once I'm happy with things, your PR will be merged! 🎉

PR #1693, opened 2019-11-14
add action to check project changes and provide feedback on each PR

shiftkey:
This PR adds an action that runs on every `pull_request` event and does a few things:

 - check the PR adds or modifies any files under `_data/projects`
 - parses the files and checks they are valid
 - for GitHub projects, run additional checks that
   - the repository exists and is valid
   - the `label` specified is in use
   - the `label` and `link` are in sync when the project links directly to the label

This is codifying several checks that I've been manually doing up until now, and it'd be great to not have to do this to save time. Ideally this comment will help contributors fix things without needing to dig into the Actions logs to see why something failed...

### DIFF
--- a/.github/actions/project-analyzer/Dockerfile
+++ b/.github/actions/project-analyzer/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:2.6-stretch
+
+# Labels for GitHub to read the action
+LABEL "com.github.actions.name"="Update project stats"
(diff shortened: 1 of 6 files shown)


shiftkey:
I've been poking at this locally with some contrived scenarios, but I'm going to merge this in next and see if I can run it against some open PRs and get some real-world testing out of it.

PR #1724, opened 2019-11-26
Create testing_shiftbot.yml

ritwik12:


ritwik12:
@shiftkey That works pretty good. How about adding one more thing.
The most common issue we have with new PRs is the one with `XYZ` label name but some `https://github.com/username/project/labels/abc` URL.
I know it will not be good to force "link:" validation as many people might have different label link other than `https://github.com/username/project/labels/` but for those who are using `https://github.com/username/project/labels`. It will be good to compare their label name and label URL to match

ritwik12:
@shiftkey You have added shiftbot to  validate only `_data/projects/`. How about checking the new file path also and if it is *.yml then to give a reply of adding it to `_data/projects/`  only

shiftkey:
@ritwik12 thanks for the feedback!

> It will be good to compare their label name and label URL to match

This is something that should be covered, and the message you see looks like this:



This is a pretty specific check - if you're a GitHub project and the existing `link` contains `/labels/` then we expect it to match the label. That felt narrow enough to not catch other cases like organizations, and I'm planning to improve the copy later.

If you want to read the code:

  - [`GitHubRepositoryLabelActiveCheck`](https://github.com/up-for-grabs/tooling/blob/master/lib/queries/github_repository_label_active_check.rb) -  this handles finding the label on GitHub
  - [`label_check` method](https://github.com/shiftkey/webhooks/blob/master/app/jobs/up_for_grabs_pull_request_project_analyzer_job.rb#L163-L196) -  that handles the result from `GitHubRepositoryLabelActiveCheck` and converts it into a user-friendly message if there's a problem

> How about checking the new file path also and if it is *.yml then to give a reply of adding it to `_data/projects/` only

I already check this in the CI validation step (the [`DirectoryValidator`](https://github.com/up-for-grabs/tooling/blob/master/lib/validators/directory.rb) that runs [here](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/scripts/validate_data_files.rb)), and before I add it to this process I'd like to extract the common things I've found back into the [`tooling`](https://github.com/up-for-grabs/tooling) library:

 - "results" - a common shape for the result of validating the project, that supports both things that are checked as part of CI (fast feedback, less context, limited API access) or this PR commenter (more detailed, suggests solutions, full API access).
 - "formatters" - both CI and the PR commenter need to convert the result into a format that makes sense for the viewer - plain text for CI, markdown for the PR commenter

Once I'm happy with those abstractions, it should become easier to flesh out the remaining gaps or relevant cases that we find ourselves identifying during reviews.

ritwik12:
@shiftkey That's great, let me know if any help/discussions needed.

PR #1734, opened 2019-12-01
Create html-duration-picker.yml

nadchif:
Added https://nadchif.github.io/html-duration-picker.js/ to listing


_data/projects/html-duration-picker.yml
```diff
@@ -0,0 +1,12 @@
+name: HTML-DURATION-PICKER.JS
```
shiftkey: This doesn't need to be all-caps, and can just be how the project is named in the README:

```suggestion
name: html-duration-picker.js
```

shiftkey (commented):
@nadchif thanks for the contribution!

Just one suggestion, because the all-caps felt distracting

_data/projects/html-duration-picker.yml
nadchif: @shiftkey thank you for the suggestion. You are right. I've applied the changes. Thanks again

shiftkey (approved):

PR #1764, opened 2019-12-23
Create adobe-brackets-encode-decode.yml

nadchif:


shiftkey (approved):
@nadchif thanks for the contribution!

PR #1806, opened 2020-01-17
Project added: htmldate

adbar:
Template used: https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/docs/list-a-project.md

This software library is now ripe and hopefully documented enough to welcome new contributors. Any help would be appreciated, thank you for your effort and the very clear instructions!


shiftkey (approved):
@adbar thanks for the contribution!

PR #1808, opened 2020-01-17
Create real-time-project.yml

wingerdkm1:
Someone just introduced our team to Up for Grabs.

Our request will be a great addition

### DIFF
--- a/_data/projects/real-time-project.yml
+++ b/_data/projects/real-time-project.yml
@@ -1,0 +1,10 @@
+name: Real Time Communications World
+desc: Maintain API Security
+site: https://www.realtimecommunicationsworld.com
+tags:
+  - api
+  - ddos
+  - json
+  upforgrabs:
+  name: up for grabs
+  link: https://github.com/username/project/labels/up%20for%20grabs


wingerdkm1 (commented):
Not sure why it says line 5 is missing "-"

_data/projects/real-time-project.yml
```diff
@@ -0,0 +1,10 @@
+name: Real Time Communications World
+desc: Maintain API Security
+site: https://www.realtimecommunicationsworld.com
+tags:
+  - api
+  - ddos
+  - json
+  upforgrabs:
```
shiftkey, line 8: I think the indents on this line and the next two need to be removed, as our parser expects an array for `tags` and these lines are being parsed as part of it.

ritwik12:
@wingerdkm1 Ping

ritwik12:
@shiftkey Did we stopped automatically closing PRs which stayed more than two weeks? I guess we used to mark them stale

shiftkey:
@ritwik12 I think #1706 was the problem. #1873 should get us back to a good place.

ritwik12:
@shiftkey Got it, Thanks!!

PR #1816, opened 2020-01-20
update mockingcase site

strdr4605:


shiftkey (approved):
Thanks for the update @strdr4605!

PR #1821, opened 2020-01-23
adding xkcd bot

joeyvanlierop:
I think that my project would be a great fit for Up for Grabs as it is very well documented and includes issues that can be approached by both beginners and advanced users!

Thanks 👋


shiftkey (approved):
Thanks for the contribution @joeyvanlierop!

PR #1842, opened 2020-02-01
React Test

franciscop:
Added library React Test

### DIFF
--- a/_data/projects/react-test.yml
+++ b/_data/projects/react-test.yml
@@ -8,6 +8,7 @@ tags:
   - jquery
   - expressive
   - act
+  - render
   - jest
   - runner
 upforgrabs:
--- a/_data/projects/react-test.yml
+++ b/_data/projects/react-test.yml
@@ -2,7 +2,7 @@ name: React Test
 desc: An expressive testing library for React
 site: http://react-test.dev/
 tags:
-  - react
+  - reactjs
   - test
   - testing
   - jquery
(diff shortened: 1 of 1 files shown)


ritwik12 (changes requested):
@franciscop Please fix given suggestions

_data/projects/react-test.yml
```diff
@@ -0,0 +1,16 @@
...
+  - testing
+  - jquery
+  - expressive
+  - act
+  - render
+  - jest
+  - runner
+upforgrabs:
+  name: good first issue
+  link: https://github.com/franciscop/react-test/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22
```
ritwik12: @franciscop Please change this to https://github.com/franciscop/react-test/labels/good%20first%20issue

_data/projects/react-test.yml
```diff
@@ -0,0 +1,16 @@
+name: React Test
```
ritwik12, line 1: @franciscop Can we change the name to something else so that users will not think that this is Official https://reactjs.org/ product.

ritwik12 (changes requested):

_data/projects/react-test.yml
franciscop: Oh I didn't think about it getting confused because it's very common to do `react-X` in the community for 3rd party packages, but of course if it's just `react-test` it could easily be confused! I'll update the project site and everything as well, thanks for the input!

ritwik12:
@franciscop Ping

franciscop:
Hi, made changes on both of the suggestions:

- Added a non-official notice in the description.
- Changed the link for the label.

ritwik12 (approved):

PR #1873, opened 2020-02-18
Revert "adding a new tag that indicates the reviewers have questions"

shiftkey:
Reverts up-for-grabs/up-for-grabs.net#1706

I misread the documentation at the time and didn't notice this comment:

> `# Only issues or pull requests with all of these labels are check if stale. Defaults to [] (disabled)`

The `all` here is the bit that I overlooked.

What I think has happened here is that it's only going to look at issues which have both labels, and because they only have one label they're being skipped from cleanup.

Let's back this out for now and see if we can figure out a better solution.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -5,7 +5,6 @@ pulls:
   daysUntilClose: 1
   onlyLabels:
     - 'needs work'
-    - 'has questions'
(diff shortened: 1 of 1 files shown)


ritwik12:
Got it @shiftkey

PR #1834, opened 2020-01-27
Add RDoc::Generator::SolarFish

gavv:
Add [RDoc::Generator::SolarFish](https://github.com/rbdoc/rdoc-generator-solarfish), single-page HTML5 generator for Ruby RDoc.

PR #1908, opened 2020-03-03
Delete rdoc-generator-solarfish.yml

gavv:
Hi,

I decided to archive the project, so deleting it from here.

It was added recently in #1834.


shiftkey (approved):
Thanks for the tidy-up @gavv!

PR #1910, opened 2020-03-03
Add pubgate-philip project

autogestion:
I think my project would be great for those who wants to try their skills with Svelte framework, or/and interested in decentralized technologies


shiftkey (approved):
Thanks for sharing @autogestion!

PR #1930, opened 2020-03-16
Add Simple Icons

PeterShaggyNoble:
I went with all the tags we use for [our GitHub repo](https://github.com/simple-icons/simple-icons/) - don't know if that's overkill.

### DIFF
--- a/_data/projects/simple-icons.yml
+++ b/_data/projects/simple-icons.yml
@@ -1,0 +1,27 @@
+name: Simple Icons
+desc: Free SVG icons for popular brands.
+site: https://simpleicons.org/
+tags:
+- brand
+- branding
+- brands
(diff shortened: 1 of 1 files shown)


shiftkey:
@PeterShaggyNoble 

> I went with all the tags we use for [our GitHub repo](https://github.com/simple-icons/simple-icons/) - don't know if that's overkill.

Probably - we typically get 5-10 tags for a project. Maybe trim it down to the important ones?

PeterShaggyNoble:
Thanks for the quick feedback, @shiftkey - I've trimmed it down to 5 key tags.

shiftkey (approved):
LGTM, thanks for the contribution @PeterShaggyNoble!

PR #1972, opened 2020-04-06
switch checkout steps to v2

shiftkey:
This PR bumps our checkout step to `checkout@v2` which has some significant improvements for large repositories. 

Not sure how much it'll benefit us, but this might handle #1967 better.


ritwik12:
@shiftkey Thanks!!

PR #2015, opened 2020-04-27
adding my new project

nadchif:
I think this project would be a good project for Up for Grabs. We will be offering guidance and assistance to first timers who would love to participate.

Please let me know if there's anything else I can do to meet the Up for Grabs requirements

### DIFF
--- a/_data/projects/headset-control-plus.yml
+++ b/_data/projects/headset-control-plus.yml
@@ -1,0 +1,14 @@
+name: Headset Control Plus
+desc: Unleash the power of your Headset button by supporting gesture like: Long Press, Double Press, Triple Press etc.
+site: https://play.google.com/store/apps/details?id=com.chif.headsetcontrolplus
+tags:
+  - android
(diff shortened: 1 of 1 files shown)


_data/projects/headset-control-plus.yml
```diff
@@ -0,0 +1,14 @@
+name: Headset Control Plus
+desc: Unleash the power of your Headset button by supporting gesture like: Long Press, Double Press, Triple Press etc.
```
shiftkey: The build error is due to the second `:` here being parsed incorrectly. I recommend tweaking the words:

```suggestion
desc: Unleash the power of your Headset button by supporting gestures such as Long Press, Double Press, Triple Press etc.
```

nadchif: Nice catch! Thank you

shiftkey (approved):
Thanks for the contribution @nadchif!

PR #2016, opened 2020-04-28
Updated after.html

sandip2224:
Grammatical fix proposed


ritwik12:
@sandip2224 I am not sure if that is a Grammatical mistake. Hence, closing this being lying over here without discussions. Please raise an issue if you believe there is a change needed. Thanks for opening the PR.

PR #2022, opened 2020-05-02
Adding Voodoo2D project

JacksonHoggard:
I am requesting to add my project to the site because I have set up the repo in a way that it is friendly and inviting to beginners! It is also an awesome project that I think is perfect for up-for-grabs. Let me know if there are any issues with the template!


shiftkey (approved):
@CremBluRay thanks for the contribution!

PR #2054, opened 2020-05-11
Update QuickPerf projects

jeanbisutti:
Hello,

I have updated the QuickPerf projects. I am the main maintainer of these projects.

Jean

### DIFF
--- a/_data/projects/maven-test-bench.yml
+++ b/_data/projects/maven-test-bench.yml
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: good first issue
+  name: : "thumbsup: good first issue"
   link: https://github.com/quick-perf/maven-test-bench/labels/%3Athumbsup%3A%20good%20first%20issue
 stats:
   issue-count: 0
--- a/_data/projects/maven-test-bench.yml
+++ b/_data/projects/maven-test-bench.yml
@@ -1,5 +1,5 @@
 ---
-name: Maven test bench
+name: Maven performance test bench
 desc: Performance test bench for Apache Maven project.
 site: https://github.com/quick-perf/maven-test-bench
 tags:
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: ":thumbsup: good first issue"
+  name: good first issue
   link: https://github.com/quick-perf/maven-test-bench/labels/%3Athumbsup%3A%20good%20first%20issue
 stats:
   issue-count: 0
--- a/_data/projects/maven-test-bench.yml
+++ b/_data/projects/maven-test-bench.yml
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: : "thumbsup: good first issue"
+  name: ":thumbsup: good first issue"
   link: https://github.com/quick-perf/maven-test-bench/labels/%3Athumbsup%3A%20good%20first%20issue
 stats:
   issue-count: 0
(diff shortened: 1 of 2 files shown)


ritwik12 (changes requested):
@jeanbisutti I guess you forgot to change label names on those repos. There it is still Thumbup good first issue

jeanbisutti:
> @jeanbisutti I guess you forgot to change label names on those repos. There it is still Thumbup good first issue

@ritwik12 , thanks for your review. I am sorry but [I sew only `name: good first issue`](https://github.com/up-for-grabs/up-for-grabs.net/pull/2054/files), as it is done in other projects

ritwik12:
@shiftbot

ritwik12:
@jeanbisutti  please update those in your repos.

shiftkey:
It is currently failing because the label name on these specific projects starts with the emoji character `:thumbsup:` for both projects:



If you look at the API response, it'll show the same text:



We don't do any fuzzy matching here, because we need to lookup through the API which requires an exact match for the label name.

ritwik12:
@shiftkey Do we have a command to rebuild @shiftbot for changes such as these. Because now there are no commits here but only name change on the owners repo.

shiftkey:
@ritwik12 validating the entire file would happen on any field change, and I still think it'll complain because of the label changes in this PR (this would be fine if those are reverted). 

Re-running the validation isn't supported yet, open an issue over on https://github.com/shiftkey/webhooks if you'd like to chat about that further.

ritwik12:
Cool. Thanks @shiftkey

_data/projects/maven-test-bench.yml
```diff
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: ":thumbsup: good first issue"
+  name: good first issue
```
shiftkey: Reverting this to it's old value is what's required here:

```suggestion
  name: ":thumbsup: good first issue"
```

_data/projects/quickperf.yml
```diff
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: ":thumbsup: good first issue"
+  name: good first issue
```
shiftkey: Reverting this to it's old value is what's required here:

```suggestion
  name: ":thumbsup: good first issue"
```

_data/projects/maven-test-bench.yml
jeanbisutti: Thanks @shiftkey

ritwik12: @jeanbisutti Here, also you will have to revert.
```suggestion
  name: : "thumbsup: good first issue"
```

ritwik12 (changes requested):

ritwik12:
@jeanbisutti Sorry my suggestion had an extra : 

please make it ":thumbsup: good first issue" from : ":thumbsup: good first issue"

_data/projects/maven-test-bench.yml
```diff
@@ -8,7 +8,7 @@ tags:
 - performance
 - quickperf
 upforgrabs:
-  name: ":thumbsup: good first issue"
+  name: : "thumbsup: good first issue"
```
ritwik12: @jeanbisutti Please remove this extra colon outside ""
```suggestion
  name: ":thumbsup: good first issue"
```

ritwik12 (changes requested):

ritwik12:
@jeanbisutti Thanks a lot!!

jeanbisutti:
Huge thanks for your help @ritwik12!

PR #2078, opened 2020-05-22
add support for opening in Codespaces

shiftkey:
Following on from #2077 I was able to get this project working in Codespaces with this bit of setup:

 - updating the `docker-compose.yml` to a later version
 - adding the `.devcontainer.json` file to use this `docker-compose` setup 

If you have access to the Codepsaces beta, here's how you see this work:

 - Under "Code" in the header of the repository, select "Open with Codespaces"

 - Find the "Remote Explorer" icon in the left sidebar

remote-containers

 - In the lower part of the left column, you'll see the Codespaces detail and the port this container has open. Click on 4000 to open the site in a new browser tab.

open-port

  - You might have to wait a moment because the container installs dependencies and builds the site at this point, but eventually you should see the Up for Grabs site hosted on Codespaces

website-on-codespaces

It'd be great to get hot reloading working properly, but you can see that changes to the projects are updated after refreshing the page

edit-and-refresh

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,5 +24,8 @@ ENV LC_ALL C.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
 
+RUN echo "Testing program is on PATH"
+RUN bash --version
+
 EXPOSE 4000
 CMD bundle install && bundle exec jekyll serve -d /_site --watch --force_polling -H 0.0.0.0 -P 4000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,8 @@
-app:
-  build: .
-  ports:
-    - '4000:4000'
-  volumes:
-    - .:/app
+version: "3"
+services:
+  app:
+    build: .
+    ports:
+      - "4000:4000"
+    volumes:
+      - ".:/app"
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -24,5 +24,8 @@ ENV LC_ALL C.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
 
+RUN echo "Testing program is on PATH"
+RUN bash --version
```
shiftkey, line 28: I encountered an issue here where Codespaces was trying to install some additional things it needed inside my container, and on Alpine it didn't have `bash` installed by default, nor did it have `/bin` on the PATH.

This is a check to ensure we provide it for this tooling.

docker-compose.yml
```diff
@@ -1,6 +1,8 @@
-app:
-  build: .
-  ports:
-    - '4000:4000'
-  volumes:
-    - .:/app
+version: "3"
```
shiftkey, line 1: Turns out `docker-compose` files now have versions, and 3.x is the latest supported version.

I caused an error here in Codespaces by using a `3.7` version and instead set it to `3`, so it all seems fine now :man_shrugging:

PR #2097, opened 2020-05-27
regenerated project stats

oliviadiazf:
I think my project would be a great fit for Up for Grabs.


shiftkey:
Looks like we've got several PRs open related to `Scikit-learn` - let's just use #2099 as that's the latest PR.

PR #2144, opened 2020-06-20
create book project entry

knjk04:
I think the Book Project is a good fit for Up for Grabs :)


shiftkey:
@knjk04 thanks for the contribution! Unfortunately this file needs to be moved out of the repository root - could you update the PR?

```
1 files found in root which look like project files:
  - book-project.yml
Move these inside _data/projects/ to ensure they are listed on the site
```

knjk04:
@shiftkey Oops, sorry about that! I've fixed it now

ritwik12:
Thanks!! @knjk04

PR #2191, opened 2020-07-10
Update Roc.yml

gavv:
The name was changed to "Roc Toolkit" and the URL was changed too.


ritwik12:
@gavv Thanks for updating it.

PR #2192, opened 2020-07-10
Add signal-estimator.yml

gavv:
Hi, signal-estimator is a small tool that we're using for testing in [Roc](https://github.com/roc-streaming/roc-toolkit), which already added to up-for-grabs.

https://github.com/gavv/signal-estimator


ritwik12:
@gavv That's cool, thanks!!

PR #2228, opened 2020-08-01
Create time-management.yml

ghost:
_time-management_  (TM) is a very simple Python program with a console based interface that I use to keep track of my daily activities as a software developer. It's useful and accessible for computer science students and new developers.

The beauty of _time-management_  is that you can use the program to keep track of your work on the program!


ghost:
Word.

ritwik12:
@ancbro Thanks!!
We are based out of different time zones so it might not be the case that there is an update on some PR in just few hrs.

PR #2239, opened 2020-08-04
Add ArviZ

OriolAbril:
We think ArviZ would be a good addition to up for grabs. 

We are active and willing to help more people get started with open source.


shiftkey (approved):
@OriolAbril thanks for the contribution!

PR #2244, opened 2020-08-07
Create xunit.ts.yml

SteveDesmond-ca:
:wave: Now that we've got our first "up for grabs" issue available, why not share in the OSS love?


shiftkey (approved):
@stevedesmond-ca thanks for the contribution!

PR #2246, opened 2020-08-08
create stumpless.yml

goatshriek:
Stumpless has continued to maintain issues for newcomers, and has recently had some of these grabbed and completed. Now that the documentation has been updated with feedback from new developers, I believe the project is ready to post here so that it is more discoverable by folks looking to learn and contribute.


shiftkey (approved):
@goatshriek thanks for the contribution!

PR #2272, opened 2020-08-26
Adding .NET nanoFramework

josesimoes:
Here's a project that is very interesting for .NET developers that also like embedded systems.

Please let me know if I need to change anything on submission.


shiftkey (approved):
Thanks for the contribution @josesimoes!

PR #2283, opened 2020-09-04
Add OpenTelemetry Projects

svrnm:
Hi,
this PR adds the open telemetry projects from https://github.com/open-telemetry/
See also the community ticket here: https://github.com/open-telemetry/community/issues/469

Thanks,


ritwik12 (commented):
@svrnm Thanks for sending in the PR. All looks good. Only one issue is there that there are some projects which do not have any issues labeled with the tag `help wanted` on their repository. Would request you to please add those so that if someone visits the project site they would not be treated with 0 issues.

PR #2328, opened 2020-09-26
Add `lxd` issues labeled `easy`

stgraber:
LXD has had issues labeled as `easy` which are great for new contributors.
We actively provide mentoring on those and have had a lot of success with students and other first time contributors using them to get familiar with Go or containers/VM management.


ritwik12:
@stgraber Thanks!!

PR #2331, opened 2020-09-27
Added our project DX Scanner

DominikMilaniak:
DX Scanner is an awesome fit for up-for-grabs, because it directly helps developers and their team leaders and is a good project to improve your typescript skills on.


ritwik12:
@DominikMilaniak Thanks!!

PR #2348, opened 2020-09-29
Correcting Label Link for Arc.yml

ritwik12:
@coolaj86 While reviewing I missed that the Label link was wrong. Correcting that with this PR.


coolaj86:
Thanks.

PR #2354, opened 2020-09-30
Update how-it-works.md

chxtan:


shiftkey:
Thanks for the interest, but documentation contributions need to be more substantial than this.

PR #2356, opened 2020-09-30
updated style sheet for better user experience

thishantj:


shiftkey:
It's not clear to me what's happening here, so I'm going to close this out. 

If you are interested in contributing, please use a better PR description and add more context about the changes you are making to the PR body so I can understand the situation better.

PR #2357, opened 2020-09-30
Update stylesheet.css

Sharuka-Rajapaksha:


ritwik12:
@Sandeepana Please share the intentions to change those color codes? What color is it and what it will affect. and why this is needed at all.

Sharuka-Rajapaksha:
Dear Sir,

I'd change that color because it was a very pleasant color for the eye and
makes us feel great.

Thank you.

shiftkey:
@Sandeepana I took at look at our staging deployment with these changes:



I'm not a fan of this colour, and the shade of green makes parts of the site harder to read:



Closing this out as I don't feel it is the right direction.

PR #2361, opened 2020-10-01
Updated the Update setup.md

JaySingh23:


ritwik12:
@JaySingh23 Thanks for your contribution but I don't think there is a need for that.

PR #2367, opened 2020-10-01
Styles has been updated to get better user interface and user experience

thishantj:
Style sheet has been updated to give better user viewing between colours and contrasts. This is to ensure that users with different impairments aren't left behind with difficulty in using the system, which ultimately increases user satisfaction.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,0 +1,318 @@
+// @ts-nocheck
+
+define([
+  'jquery',
+  'projectLoader',
+  'projectsService',
+  'fetchIssueCount',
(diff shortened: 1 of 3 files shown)


shiftkey:
Thanks for your interest, but we're currently not interested in these types of contributions.

Please check out our list of [up for grabs issues](https://github.com/up-for-grabs/up-for-grabs.net/labels/up-for-grabs) or [the site itself](https://up-for-grabs.net/) for projects looking for new contributors.

PR #2389, opened 2020-10-05
InstaGrabber => Barinsta

austinhuang0131:
change of project name due to rebrand

was #2241


ritwik12:
@austinhuang0131 Thanks for updating!!

PR #2393, opened 2020-10-07
Update DNN project file with correct information

david-poindexter:
Updates to include up-to-date tags and label used for "up for grabs"


ritwik12:
@david-poindexter Thanks!!

PR #2434, opened 2020-10-26
Create dotman.yml

Bhupesh-V:
> ### [dotman](https://github.com/Bhupesh-V/dotman) is a simple, elegant & easy to use dotfiles manager


shiftkey (approved):
Thanks for the contribution @Bhupesh-V!

PR #2467, opened 2020-11-17
Adding my mjml-pardot project

Alexnortung:
Adding my mjml-pardot project,
this project has a few issues that are good for new contributers.


shiftkey (approved):
Thanks for the contribution @Alexnortung!

PR #2475, opened 2020-11-22
adding my new project

skekre98:
👋 Hi I think my project would be a good fit for Up For Grabs. I have maintained other entry level open source projects in the past as well. Let me know if you have any feedback on the template.


shiftkey (approved):
Thanks for the contribution @skekre98!

PR #2538, opened 2021-01-08
Add Smack (XMPP client library)

Flowdalic:


_data/projects/Smack.yml
```diff
@@ -0,0 +1,13 @@
...
+  Smack is an open source, highly modular, easy to use, XMPP client library written in Java for Java SE compatible JVMs and Android.
+site: https://www.igniterealtime.org/projects/smack/
+tags:
+- xmpp
+- java
+- im
+- client
+upforgrabs:
+  name: Teaser Tasks
+  link: https://igniterealtime.atlassian.net/issues/?filter=10003
```
shiftkey, line 13: Not a blocker on merging, but when I follow this link I get this error:

`The requested filter doesn't exist or is private. Try logging in.`

Does that sound right?

Flowdalic: The JIRA filter should be public now.

shiftkey: Still no change on my end 🤔

Flowdalic: Now it should work. Sorry.

shiftkey (approved):
Thanks for the contribution @Flowdalic!

PR #2554, opened 2021-01-24
Added new project.

Sh-wayz:
Just found out about this and figured I would add an open source project I've been contributing to for a while, let me know if there's anything I need to fix!


shiftkey (approved):
Thanks for the contribution @Sh-wayz!

PR #2561, opened 2021-01-28
A Midnight Commander fork – neo-mc

psprint:
I'm looking for people to work on a Midnight commander fork called neo-mc (⋱הϵѻ⋱ Midnight Commander). To give a justification of why to join: The code is very good written and VERY entertaining, for example there are Widgets structs defined with the C-simulated class hierarchy (`struct WListbox { Widget base; int extending_field; … }`) or GLib's data structures used (single/double linked lists, hash tables, quarks, etc.). It is very fun to work on the code.

The goal of the fork is to make `mcedit` competitive with Vim&Emacs, by addition of a scripting language and other improvements. I have already merged 7 feature branches. But there is much that [can be done](https://github.com/neo-mc/neo-mc/wiki/MCEditWishList). The project is at: https://github.com/neo-mc/neo-mc.

### DIFF
--- a/_data/projects/neo-mc.yml
+++ b/_data/projects/neo-mc.yml
@@ -12,5 +12,4 @@ upforgrabs:
   link: https://github.com/neo-mc/neo-mc/wiki/MCEditWishList
 stats:
   issue-count: 0
-  last-updated: '2021-01-27T23:15:00Z'
   
--- a/_data/projects/neo-mc.yml
+++ b/_data/projects/neo-mc.yml
@@ -1,0 +1,16 @@
+---
+name: neo-mc
+desc: Next generation fork of Midnight Commander
+site: https://github.com/neo-mc/neo-mc
+tags:
+- c
+- glib
+- slang
+- cli
+upforgrabs:
+  name: wish-list
+  link: https://github.com/neo-mc/neo-mc/wiki/MCEditWishList
+stats:
+  issue-count: 0
+  last-updated: '2021-01-27T23:15:00Z'
+  


_data/projects/neo-mc.yml
```diff
@@ -0,0 +1,16 @@
...
+- c
+- glib
+- slang
+- cli
+upforgrabs:
+  name: wish-list
+  link: https://github.com/neo-mc/neo-mc/wiki/MCEditWishList
+stats:
+  issue-count: 0
+  last-updated: '2021-01-27T23:15:00Z'
```
shiftkey: Dropping this line should address the listing issue (if there are no open issues, it's unclear when they were last updated):

```suggestion
```

_data/projects/neo-mc.yml
```diff
@@ -0,0 +1,15 @@
...
+desc: Next generation fork of Midnight Commander
+site: https://github.com/neo-mc/neo-mc
+tags:
+- c
+- glib
+- slang
+- cli
+upforgrabs:
+  name: wish-list
+  link: https://github.com/neo-mc/neo-mc/wiki/MCEditWishList
```
shiftkey, line 12: I think this is the first time in my many years of reviewing project submissions that a wiki page has been used. I think this severely breaks the automation I've used for GitHub projects so it's gonna need some work to resolve this and get the PR passing.

If you're set on doing it this way, rather than using issues in the repository, please let us know.

ritwik12: @psprint ^^

PR #2616, opened 2021-03-09
Added auth-jwt

gitcommitshow:
A project to teach about JWT by reverse engineering

### DIFF
--- a/_data/projects/auth-jwt.yml
+++ b/_data/projects/auth-jwt.yml
@@ -1,0 +1,13 @@
+---
+name: auth-jwt
+desc: 'A demo to learn JWT by reverse engineering'
+site: https://github.com/gitcommitshow/auth-jwt
+tags:
+- jwt
+- authentication
+- learning
+- rfc-7519
+- express
+upforgrabs:
+  name: good first issue
+  link: https://github.com/gitcommitshow/auth-jwt/labels/good%20first%20issue
--- a/_data/projects/auth-jwt.yml
+++ b/_data/projects/auth-jwt.yml
@@ -1,6 +1,6 @@
 ---
 name: auth-jwt
-desc: 'A demo to learn JWT by reverse engineering'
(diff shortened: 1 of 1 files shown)


_data/projects/auth-jwt.yml
```diff
@@ -0,0 +1,13 @@
+---
+name: auth-jwt
+desc: 'A demo to learn JWT by reverse engineering'
```
ritwik12: ' is not needed here.
```suggestion
desc: A demo to learn JWT by reverse engineering
```

_data/projects/auth-jwt.yml
```diff
@@ -0,0 +1,13 @@
...
+site: https://github.com/gitcommitshow/auth-jwt
+tags:
+- jwt
+- authentication
+- learning
+- rfc-7519
+- express
+upforgrabs:
+  name: good first issue
+  link: https://github.com/gitcommitshow/auth-jwt/labels/good%20first%20issue
```
ritwik12, line 13: You don't have any issues marked with this label, at least mark few issues so that users can visit and work on that.

ritwik12 (changes requested):

_data/projects/auth-jwt.yml
gitcommitshow: Fixed in latest commit. Can you please review again.

_data/projects/auth-jwt.yml
gitcommitshow: Added [issue with `good first issue`](https://github.com/gitcommitshow/auth-jwt/issues/6) tag. Will keep adding more.
Is it good to go now?

ritwik12: Yep

_data/projects/auth-jwt.yml
ritwik12: Ohk

PR #2618, opened 2021-03-13
Update nlog.yml

snakefoot:
Removed obsolete platforms silverlight and mono (now netcore) and instead added structured + semantic

See also: https://github.com/NLog/NLog/wiki/How-to-use-structured-logging


ritwik12:
@snakefoot Thanks!!

PR #2619, opened 2021-03-13
Update nlog.yml

snakefoot:
Removed Silverlight from the description as well.


ritwik12:
@snakefoot Thanks for updating it.

PR #2642, opened 2021-03-29
Add sigstore/{rekor, cosign, fulcio} projects

efebarlas:
@dlorenc @SantiagoTorres 

We would like to add three projects. Some of these projects do not have a lot of good first issue tags but we will add more this week.


shiftkey (approved):
Thanks for the submissions @efebarlas!

PR #2643, opened 2021-03-29
Added ASP.NET Core to the .NET party

davidfowl:
I can't believe nobody added ASP.NET 😢


shiftkey (approved):
Seems legit. Thanks for the submission @davidfowl!

PR #2648, opened 2021-03-29
Added Magicodes.IE

hueifeng:
Magicodes.IE is An import and export general library, support Dto import and export and dynamic export, support Excel, Csv, Word, Pdf and Html.


shiftkey (approved):
Thanks for the contribution @hueifeng!

PR #2678, opened 2021-04-12
Create ideahub.yml

ming-tsai:


ming-tsai:
Hi @ritwik12,
I don't know which is my error

_data/projects/ideahub.yml
```diff
@@ -0,0 +1,12 @@
+name: IdeaHub
+desc: A collection of ideas and projects that contain from beginner to advance :octocat: 🎯🚀
```
shiftkey: I think the `:` part of `:octocat:` here is not being handled correctly. If you wrap this in quote marks it should be handled as a string:

```suggestion
desc: "A collection of ideas and projects that contain from beginner to advance :octocat: 🎯🚀"
```

ming-tsai: Thanks @shiftkey

PR #2680, opened 2021-04-15
add CoFeed-19

strdr4605:
I think https://github.com/cofeed-19/cofeed-19.github.io will be a good opportunity for new contributors.
I already guided some first time contributors on other project.


ritwik12:
@strdr4605 Thanks!!

PR #2699, opened 2021-04-29
Add flutter app for wger

rolandgeider:
Add flutter app


shiftkey (approved):
Thanks for the submission @rolandgeider!

PR #2767, opened 2021-06-15
Add Class Link Check

Arihant25:
Yo 👋 this is a simple extension of mine that I would like to be added to the list. Thanks a lot!


shiftkey (approved):
Thanks for the submission @Arihant25!

PR #2870, opened 2021-09-08
Create RocDroid.yml

gavv:
Hi! This PR adds "Roc Droid", and Android app for streaming audio using Roc. Part of https://roc-streaming.org/.


ritwik12:
@gavv Thanks!!

PR #2873, opened 2021-09-08
Add Codename One

yo-zubair:
With several issues open for new contributors, Codename One would be a good fit for Up for Grabs.


ritwik12:
@yo-zubair Thanks!!

PR #2888, opened 2021-09-17
Add the Openverse project

zackkrida:
Openverse is a search engine for openly-licensed media, part of the WordPress ecosystem.

As the project is comprised of 4 different repositories, I've used the following GitHub search result as the `upforgrabs` link. Please let me know if that is acceptable.

### DIFF
--- a/_data/projects/Openverse.yml
+++ b/_data/projects/Openverse.yml
@@ -14,4 +14,4 @@ tags:
 - airflow
 upforgrabs:
   name: help wanted
-  link: https://github.com/search?q=label%3A%22good+first+issue%22+state%3Aopen+is%3Aissue+repo%3Awordpress%2Fopenverse+repo%3Awordpress%2Fopenverse-frontend+repo%3Awordpress%2Fopenverse-api+repo%3Awordpress%2Fopenverse-catalog+label%3A%22help+wanted%22+state%3Aopen+repo%3Awordpress%2Fopenverse&type=Issues
+  link: https://github.com/wordpress/openverse/labels/help%20wanted
--- a/_data/projects/Openverse.yml
+++ b/_data/projects/Openverse.yml
@@ -14,4 +14,4 @@ tags:
 - airflow
 upforgrabs:
   name: help wanted
-  link: https://github.com/wordpress/openverse/labels/help%20wanted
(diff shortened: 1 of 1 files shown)


ritwik12:
@zackkrida With that search link, we wont be able to update project stats for this project due to which we wont be able to show the number of open issues and last updated. Would be better if you can share the Github project link and issues link.

_data/projects/Openverse.yml
```diff
@@ -0,0 +1,17 @@
...
+- postgres
+- django
+- sqlalchemy
+- javascript
+- typescript
+- vue
+- airflow
+upforgrabs:
+  name: help wanted
+  link: https://github.com/search?q=label%3A%22good+first+issue%22+state%3Aopen+is%3Aissue+repo%3Awordpress%2Fopenverse+repo%3Awordpress%2Fopenverse-frontend+repo%3Awordpress%2Fopenverse-api+repo%3Awordpress%2Fopenverse-catalog+label%3A%22help+wanted%22+state%3Aopen+repo%3Awordpress%2Fopenverse&type=Issues
```
zackkrida:
```suggestion
  link: https://github.com/wordpress/openverse/labels/help%20wanted
```

zackkrida:
Thanks @ritwik12 I've switched to the `help wanted` issues link. Not sure where I would add the shared project board for the repositories, but it is https://github.com/orgs/WordPress/projects/3

ritwik12:
@zackkrida If you want to keep the shared project board, that is for the whole WordPress Org. you can either create a new PR for adding the Org with that Project board link or maybe update this one with the link. that's on you.

Three options from my side:

1. **Keep it this way only:** for that please change the link to `https://github.com/WordPress/openverse/labels/help%20wanted`. It means the same but our bots are case sensitive so it will give errors while updating the project stats.
2. **Change the issue link to Project Board:** Keep the site link the same and change the issue link.
3. **Change both site link and issue link:** Either create a separate PR or do this in this one. Change the site link to the WordPress Org link and change the issue link to the project board link.

For 2 and 3, our bot will not be able to update the project stats but we are open to any options which is more useful to you. Please let me know if any doubts.

ritwik12 (changes requested):
Please update PR with your best suitable option as given.

ritwik12 (approved):

PR #2898, opened 2021-09-22
Adding manim

Darylgolden:
Manim is an animation engine for explanatory math videos. We have a welcoming community and are more than willing to mentor new contributors.


ritwik12:
@Darylgolden Thanks!!

PR #2933, opened 2021-10-05
adding my new project — pptr.io

choraria:
pptr.io — A free and open-source api that runs puppeteer as a service


shiftkey (approved):
Thanks for the submission @choraria!

PR #2934, opened 2021-10-05
chore: add setnemo/autogenerated-messages

setnemo:
I guess my php package for Laravel validation would be a great fit for Up for Grabs


shiftkey (approved):
Thanks for the submission @setnemo!

PR #2960, opened 2021-10-20
Update proot issue label

oxr463:
See: https://github.com/proot-me/blog/issues/20


shiftkey (approved):
Thanks for the update @oxr463!

PR #2999, opened 2021-11-30
Remove karrot

nicksellen:
We can't really manage onboarding the random new contributors that come via up-for-grabs.


shiftkey:
Not a problem, thanks for the listing!

shiftkey (approved):

PR #3023, opened 2021-12-20
Improving frontend

ghost:
See #930

Making it look like this: https://user-images.githubusercontent.com/61570792/146810927-ebb2a25c-4dfb-487a-9757-e0c08b277c7e.png

~~My changes currently look like this:~~ (they've been slightly modified actually, see comments)

image

TODO:
- [x] Get base done
- [ ] Determine if the users prefer a light theme or a dark theme so that we can determine what a good default will be
- [ ] Add light theme
- [x] Fix sizing for the sidebar so that it looks more like the concept
- [ ] Make the chosen.js dropdown look more like the concept
- [x] #3026 
- [ ] Remove unused CSS, clean up CSS
- [ ] Re-add the "back to top" button
- [x] Add [this plugin](https://github.com/jekyll/jekyll-seo-tag) so that we don't have to do all the meta tags ourselves.


ritwik12:
@ConnorAhern This looks good, few things:

- It would be better to keep the default as light mode only, our site looks much better in that.
- We can surely work on the logo, I will create a separate issue for that.
- We can discuss about that in #930

ghost:
Alright. I will update it once I get home.

ghost:
Forgot to mention it, but I have updated it.

ghost:
Oops-
image
Fixing in a second

ghost:
What exactly is going on here...? I'm not sure what it means by "exited with error code 1" - also, I tested it in my browser, and it seems to work.
image

_config.yml
```diff
@@ -27,6 +25,14 @@ exclude:
 
 github: [metadata]
 
+plugins:
+  - jekyll-seo-tag
```
ritwik12, line 29: @ConnorAhern Does this work with gh-pages? That could be the reason for failing. Read [here](https://stackoverflow.com/questions/53638446/jekyll-seo-tag-doesnt-work-within-github-pages)

ghost: It does work with GH Pages. I have used it before.

According to the StackOverflow post, they just used it wrong.

ghost:
image

ritwik12:
@ConnorAhern Working on this?

PR #3040, opened 2022-01-04
Add Ockam

mrinalwadhwa:
I think Ockam would be a great addition to the list.

We add new good first issues every week and offer mentorship for new contributors
https://github.com/ockam-network/ockam/discussions/1081


shiftkey (approved):
Thanks for the submission!

PR #3060, opened 2022-02-02
[WIP] Added default theme save functionality

### LINKED ISSUE #3044: Save dark mode in local storage
The dark mode once enabled/disabled can be stored in local storage for future reference so that the next it automatically loads in the previous mode?

developer-diganta:
Closes #3044 
I have added the function to save the selected theme so that the next time the user opens the page, it loads in the previous selected theme.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -91,6 +91,30 @@
   var rs = getComputedStyle(r) ;
   let root = document.documentElement;
   var condition=true ;
+  window.onload = check_mode();
+  function check_mode()
+  {
+    console.log("@!#241421")
+    let viewModeValue = localStorage.getItem("view-mode");
+    if(viewModeValue === "dark"){
+      root.style.setProperty("--body-back","#111111")
+      root.style.setProperty("--body-color","#eeeded")
+      root.style.setProperty("--abs","rgb(37, 37, 37)")
+      condition=false ;
+      document.getElementById("view-mode").src="/images/sun-light.png" ;
+      document.getElementById("view-mode-a").title="light-mode";
+      localStorage.setItem("view-mode","dark");
+    }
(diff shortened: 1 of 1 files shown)


ritwik12:
@developer-diganta I checked the site and it is working well. One question. Can we make this as an option? whenever someone clicks on dark/light mode option, we give them a option there saying "Keep lights on from now" or "Keep lights off from now"? something like that and once a person clicks on that, then only we save their default.

developer-diganta:
Yeah that could be done.. But should it happen with every click?

ritwik12:
@developer-diganta Every. click wont be good, how about a toggle button to save the state?

developer-diganta:
The toggle button should be below the theme change button or you know we could have a sort of pop up message at the beginning itself?

ritwik12:
@developer-diganta Pop ups are generally not good. Personally, I don't like pop up websites as not everyone is there for Dark or Light mode. Dark and Light mode is for those who are return users. So it would not be too much visible.

So by default it is light mode for us, we can do something like. Once a user clicks on Dark mode button, then only we display a small msg there below the button or near it somewhere saying "Want to keep the lights off from now? TOGGLE BUTTON" You can rephrase this as suited and best. There we can have a toggle button too as I wrote there.

or any other ideas you might have?

developer-diganta:
Yeah this sounds good... I'll then design a toggle below the theme selector... And then if you recommend any changes that can be done then?

ritwik12:
@developer-diganta Great, let me know if any help needed.

shiftkey:
Closing this out as https://github.com/up-for-grabs/up-for-grabs.net/pull/3122 has been merged and supports storing this value, but additional refinements are welcome!

PR #3075, opened 2022-02-12
Use automation account for stats workflow

shiftkey:
This change is related to how workflows from the `github-actions` token are no longer allowed to run other workflows.

I've switched this over to a token belonging to @shiftbot which I've meant to do for a while, but also disabled `APPLY_CHANGES` to prevent it from submitting PR. Hopefully we can get it back into a good place and figure out what happened with https://github.com/up-for-grabs/up-for-grabs.net/pull/3054


ritwik12:
Oh yeah, that disastrous PR.

PR #3089, opened 2022-02-13
remove System.IO.Abstractions as inactive project

shiftkey:
https://github.com/TestableIO/System.IO.Abstractions/issues/781


fgreinacher:
Hey @shiftkey, just noticed this and wanted to let you know that I am still actively maintaining SIOA, just looking for a bit more support. Should I send a new PR or do you prefer reverting this one? Cheers

shiftkey:
@fgreinacher sure - the only thing I'll mention here is that the `'state: up for grabs'` label doesn't seem to exist any more (that's how I dug into this and spotted that issue). 

But if you've got a list of issues tagged with a different label that you still consider suitable for this site a new PR that points there that is totally welcome.

PR #3103, opened 2022-02-21
Add Prayer Times App

b0bdN:
Hi! I'd like to reference my new app in your website :)

Let me know if I need to change something :)


shiftkey (approved):
Thanks for the submission @b0bdN!

PR #3113, opened 2022-02-28
Re-add System.IO.Abstractions

fgreinacher:
This reverts #3089 and also updates the up-for-grabs label.

/cc @shiftkey


shiftkey (approved):
Thanks for the submission @fgreinacher!

PR #3122, opened 2022-03-03
remember dark mode setting for returning users

### LINKED ISSUE #3044: Save dark mode in local storage

shiftkey:
Closes https://github.com/up-for-grabs/up-for-grabs.net/issues/3044
Supersedes #3060

The discussion in #3060 indicates that we want some sort of "confirm" process but I don't have time to go all that way so I'm gonna leave this open for feedback in the meantime...

### DIFF
--- a/javascripts/dark-mode.js
+++ b/javascripts/dark-mode.js
@@ -6,7 +6,44 @@ if (typeof define !== 'function') {
 }
 
 define([], () => {
-  let condition = true;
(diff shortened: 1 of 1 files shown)


ritwik12:
@developer-diganta We can have discussions if we want to add something to it.

PR #3145, opened 2022-03-25
Presenting New Look

Jai2305:
# In response to issue Improving Frontend #930
### Presenting new look for the website , 
- [x] Responsive
- [x] Compatibility with dark mode
- [x] Better use of space

- ### Light Mode
image

- ### Dark Mode
image


Jai2305:
So I think the checks are failing because of the merge conflicts , the previous version of the code used table tags , I changed them to div  for better styling , I request the developers to view the files changed section for further insights , also , ideas about the other checks and why they are failing will be very much appreciated as I am not able to understand all of them , nothing should break as I just changed html and css file that's all .

Jai2305:
One possible solution is that we can create a new branch and deploy it and test whether the code is working perfectly or not on every device and browser , I think making a pull request against gh-pages branch caused many difference .

PR #3212, opened 2022-05-18
Add Apache DataFu

eyala:
I used the template, let me know if anything is wrong. I know there's only one issue listed now but more are on their way, I promise!


datafu.yml
```diff
@@ -0,0 +1,16 @@
...
+- pig
+- java
+- python
+- spark
+- hadoop
+- python
+- hive
+upforgrabs:
+  name: up for grabs
+  link: https://issues.apache.org/jira/issues/?jql=project%20%3D%20datafu%20AND%20status%20%3D%20Open%20AND%20labels%20%3D%20up-for-grabs%20ORDER%20BY%20created%20DESC
```
ritwik12: @eyala This seems good too but we generate projects stats. So how about using `https://github.com/apache/datafu/labels/up%20for%20grabs`?
You will have to create a label `up for grabs` under that Github repo first.

ritwik12 (changes requested):

datafu.yml
eyala: The thing is, we put bugs and small feature requests in our JIRA, not on GitHub. Is it possible to add more than one link?

ritwik12: Nope, in that case I would keep this link. But then you won't be able to get projects stats on the site like other projects.

ritwik12 (approved):

ritwik12:
@eyala Please move your file to `_data/projects/`.

ritwik12:
@eyala Thanks!!

PR #3234, opened 2022-06-14
Adding Multy Project

danvdb18:
I think Multy could be an interesting project for those wanting to get more in-depth knowledge on multi-cloud / Go and would be a good fit for Up for Grabs.

Happy to hear feedback!


_data/projects/multy.yml
```diff
@@ -0,0 +1,14 @@
...
+  - go
+  - cloud
+  - aws
+  - azure
+  - terraform
+  - infrastructure-as-code
+  - multi-cloud
+upforgrabs:
+  name: good first issue
+  link: https://github.com/multycloud/multy/issues
```
ritwik12, line 14: @danvdb18 This should be https://github.com/multycloud/multy/labels/good%20first%20issue

ritwik12 (changes requested):

PR #3237, opened 2022-06-16
first try to contribute using TwilioQuest game

Faridah94:
didn't make any real changes or solve any issue I just want to practice!!


ritwik12:
@Faridah94 Thanks for sending PR but the current one looks better. 

If you just want to practice. Please pick up some projects from up-for-grabs.net and start working on those. We also have few issues open under this repo too.

PR #3248, opened 2022-06-21
Create focallocal2.yml

AndyatFocallocal:
An Opensource community platform which brings people together to take action on causes they care about, like homelessness, climate change, kindness and general well being.


AndyatFocallocal:
Is this request correct, or something still needs to be edited?

ritwik12:
> Is this request correct, or something still needs to be edited?

This should be fine. Apologies for the delay.

PR #3264, opened 2022-07-09
Mellium has moved to Codeberg

SamWhited:
The Mellium project is in the process of moving to Codeberg and just moved the issue tracker over.


ritwik12:
@SamWhited Thanks for the update!!

PR #3270, opened 2022-07-12
Added the option to filter by last updated-date

Adrianfss:
I felt that there were too many projects that were abandoned or not maintained, so I created a filter on the "last-updated" stat on each project. The buttons go from 1 week all the way to 2 years back in time.

Note that a few projects dont have the "last-updated" stat, these projects will return in the filter. 

this pull request is also related to issue #2534
Up-for-grabs-DateFilter
 

Any feedback will be much appreciated :)


ritwik12:
@Adrianfss Thanks for creating this. this was a much needed feature.

PR #3320, opened 2022-08-19
Fix dotnet/runtime

danmoseley:


danmoseley:
I removed core-clr from the doc as it no longer accepts contributions (the yml was already deleted) -- it's code moved to runtime.

ritwik12:
@danmoseley Thanks for updating it.

PR #3330, opened 2022-08-23
Update dotnet-winforms.yml

RussKie:
Use "help wanted" label instead

Similar to https://github.com/up-for-grabs/up-for-grabs.net/pull/3320

### DIFF
--- a/_data/projects/dotnet-winforms.yml
+++ b/_data/projects/dotnet-winforms.yml
@@ -1,5 +1,5 @@
 ---
-name: ".NET Core Windows Forms"
+name: ".NET Windows Forms"
 desc: >-
   Windows Forms also provides one of the most productive ways to create desktop applications based on
   the visual designer provided in Visual Studio. It enables drag-and-drop of visual controls and other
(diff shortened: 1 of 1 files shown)


RussKie:
> The label 'help wanted' for GitHub repository 'dotnet/winforms' does not match the specified `upforgrabs.link` value. Please update it to ``.

May I please get some guidance here? I've done it the same way as other configs, e.g.:
https://github.com/up-for-grabs/up-for-grabs.net/blob/96aa4aa65bf2e5b660c0eb4e1b37b78739bedf70/_data/projects/BotKube.yml#L16-L18

shiftkey:
@RussKie I think this is just a bug with the automation, based on the "empty string" value here:

> The label 'help wanted' for GitHub repository 'dotnet/winforms' does not match the specified `upforgrabs.link` value. Please update it to ``.

I've manually tested and this should be good to go.

shiftkey (approved):

RussKie:
Thank you!

PR #3346, opened 2022-08-29
Fix issue queries to satisfy bot

gewarren:


ritwik12:
@gewarren Bot was happy with that. We use`https://github.com/dotnet/docs/labels/label_name` to update stats in all the project files such as #3345

shiftkey:
Yep, the existing URLs are valid. Let's move this over to #3382

PR #3362, opened 2022-09-12
Add H2O Wave

mturoci:
H2O Wave is accepting contributions :)


mturoci:
@ritwik12 seems like the `upforgrabs.links` is correct despite bot not liking it. Can you advise on what needs to be done or is this good to be merged as is? Thanks!

ritwik12:
@mturoci That's some bug with automation. let me merge it

PR #3363, opened 2022-09-12
Update dotnet-docs-desktop.yml

adegeo:
Changing label


adegeo:
Hmm... Not sure what to do here to fix the problem.

ritwik12:
@adegeo That's some bug with automation. ignore that

PR #3396, opened 2022-09-26
Adding .net to the list of tags for Material Design in XAML

Keboo:
I am the current maintainer of this library.


shiftkey (approved):
Thanks for the contribution!

Issue #3406, opened 2022-09-29
Welcome Hacktoberfest 2022 Visitors!


ritwik12:
Welcome, all participants. Thanks for visiting up-for-grabs.net

Please go through [README.md](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/README.md), [Instructions to open a pull request](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/docs/list-a-project.md) and [CONTRIBUTING.md](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/.github/CONTRIBUTING.md)

These few links should answer all your doubts, If you still need help or face any issues please feel free to use this ISSUE as a Discussion where you can get help from other awesome contributors.

Please go through [Hacketoberfest Pull/Merge Request Details](https://hacktoberfest.com/participation/#pr-mr-details) to contribute better and follow the contributing guidelines. This would make our work much easier. I hope we will not mark even a single Pull request as "invalid". Let's work together to reduce spam.

Few Tips
For most of you, this might be the first time contributing to Open Source. Wish you all lots of new learnings in this journey ahead. I hope some of you will find their interests and would help to make Open Source a better place for people to learn and contribute. In the beginning, most of you might find it tiring to get a good repository to contribute. Please hang on to that and you will surely get something where your skills will match. Use this time to Motivate yourself and give your best to do something new and Innovate. Help others whenever you can.

Me(@ritwik12) and @shiftkey will try our best to answer you and take a look at your PRs ASAP. Please be patient and we will be right there soon. We are happy for you all to be here and find projects that will help you in contributing to Hacktoberfest and eventually start your Open Source Journey.

## About up-for-grabs
[README.md](.github/README.md)

## Guidelines 
- No spam.
- PRs for Hacktober will only be considered if an issue was raised beforehand.
- Maintain proper code of conduct.
- Be polite.
- Publishing others' private information, such as a physical or electronic address, without explicit permission will no be tolerated.
- Showing empathy towards other community members.
- Being respectful of differing viewpoints and experiences.

**SPAMMY PULL/MERGE REQUESTS WILL BE LABELED AS “SPAM.”**
We won't hesitate in labelling spammy requests as “spam” and close them. PR/MRs labeled “spam” won’t count toward Hacktoberfest. Contributors with 2+ spammy PR/MRs are disqualified.

[...]

PR #3407, opened 2022-09-29
Creating a README for Hacktoberfest.

SaadOps:


ritwik12:
@SaadAnsari17 I think we should be good with just an issue #3406 about it as I have already created. We are happy to welcome contributors but we don't want this repo to be Hacktoberfest specific. Having a separate readme about is not needed.

Thanks for creating it though, I will pick some details from it and add it in #3406

SaadOps:
> @SaadAnsari17 I think we should be good with just an issue #3406 about it as I have already created. We are happy to welcome contributors but we don't want this repo to be Hacktoberfest specific. Having a separate readme about is not needed.
> 
> Thanks for creating it though, I will pick some details from it and add it in #3406

Okay 👍 :)

PR #3441, opened 2022-10-03
Adding stats: issue-count in yml files

SaadOps:


shiftkey:
I think the issue here is that we have some `.yaml` files that aren't being updated automatically:

https://github.com/up-for-grabs/tooling/blob/abd9e70a8adc0e75671a1732eb3eddee84fd0001/lib/models/project.rb#L39-L42

```rb
    Dir["#{root}/_data/projects/*.yml"].map do |f|
      relative_path = Pathname.new(f).relative_path_from(root).to_s
      Project.new(relative_path, f)
    end
```

I think an easier fix here is to just rename those files (without changing the contents) and we can re-run the automation and regenerate the right data for free. Does that interest you @SaadAnsari17 ?

_data/projects/rawsec-inventory.yml
```diff
@@ -14,3 +14,5 @@ tags:
 upforgrabs:
   name: difficulty::easy
   link: https://gitlab.com/rawsec/rawsec-cybersecurity-list/-/issues?label_name%5B%5D=difficulty%3A%3Aeasy
+stats:
+  issue-count: 22
```
shiftkey, line 18: We don't currently support reading project details from non-GitHub hosts, which is why these details are not populated currently...

SaadOps:
> I think the issue here is that we have some `.yaml` files that aren't being updated automatically:
> 
> https://github.com/up-for-grabs/tooling/blob/abd9e70a8adc0e75671a1732eb3eddee84fd0001/lib/models/project.rb#L39-L42
> 
> ```ruby
>     Dir["#{root}/_data/projects/*.yml"].map do |f|
>       relative_path = Pathname.new(f).relative_path_from(root).to_s
>       Project.new(relative_path, f)
>     end
> ```
> 
> I think an easier fix here is to just rename those files (without changing the contents) and we can re-run the automation and regenerate the right data for free. Does that interest you @SaadAnsari17 ?

Yeah Sure !

shiftkey:
Closing this out as I don't believe this is the right approach

AbsolytnaiaLove:
сб, 8 жовт. 2022, 17:42 користувач Brendan Forster ***@***.***>
пише:

AbsolytnaiaLove:
сб, 8 жовт. 2022, 23:36 користувач AbsolytnaiaLove ***@***.***>
пише:

PR #3446, opened 2022-10-03
Fix Fulcio issue link

prateeknischal:
Fix Fulcio project's issue link which should also fix the incorrect number of issues reporting in the badge.


shiftkey (approved):
Thanks for the update @prateeknischal!

PR #3493, opened 2022-10-13
Add Bitwarden clients repository

djsmith85:


_data/projects/bitwarden-clients.yml
```diff
@@ -0,0 +1,43 @@
+---
+name: Bitwarden Clients
+desc: >-
+  Bitwarden client applications (web, browser extension, desktop, and cli).
```
shiftkey: This field is too long, and I'm honestly not sure it won't break the site. For reference, most projects only fill out a sentence or two here:

shiftkey: Nevermind, you can see it on staging:



Can we shrink this down a bit?

djsmith85:
@shiftkey Thank you for reviewing this. I agree I was trying to cram a lot of information in there. I have shortened the description and it should look good now.

shiftkey (approved):

PR #3517, opened 2022-10-19
Update image with current gui setup

VickiLanger:
Figured I'd update the image in `list-a-project`


shiftkey (approved):
Oh wow, that is an old picture. Thanks for the update!

PR #3520, opened 2022-10-20
Re-aligning the main ribbon in the header to the centre

divyanash911:
The ribbon is in the extreme right which can be aligned to the centre by simply changing the value of right attribute
Before:
image
After:
image


divyanash911:
@shiftkey if there are no modifications / addition requires please merge the PR

ritwik12:
@shiftkey Please merge if looks good to you

stylesheets/stylesheet.css
```diff
@@ -704,7 +704,7 @@ form {
   text-align: center;
   padding: .5em 1em;
   margin: 0 auto 0;
-  right: 2em;
+  right: 15em;
```
shiftkey, line 707: This breaks viewing the header when the viewport is narrow, because it's a hard-coded value:

Screenshot 2023-08-27 at 4 29 35 PM

I'd recommend looking for a solution that works better at different widths (flexbox?), and considering how the header looks when testing in responsive mode in your local browser dev tools:

Screenshot 2023-08-27 at 4 27 58 PM

shiftkey (changes requested):

shiftkey:
Closing this out for now as I don't believe this is the right approach. See previous thread for scenario to test, and ensure the approach works well when the viewport is narrow.

PR #3522, opened 2022-10-21
Removing the background behind main headings

divyanash911:
The background behind main headings in the website seems a bit dull and removing it makes the content blocks look neater. However this is subjected to the personal opinions/taste of the authorisers.
Before - 
image

After-
image


ritwik12:
@dp1136 Background looks cool, thanks for your suggestions :)

PR #3563, opened 2022-11-01
just removed the HTML element and made clean

milanshresta:
[just removed the HTML element and made clean](https://github.com/up-for-grabs/up-for-grabs.net/commit/ec5f174d51c0e06628c6cf6b8dd5c9ae708d6dca)


stylesheets/stylesheet.css
```diff
@@ -41,87 +41,9 @@ MeyerWeb Reset
...
-ruby,
-section,
-summary,
-time,
-mark,
-audio,
-video {
+*,
+::before,
+::after {
```
shiftkey: 🤔 are there any downsides to including these psuedo-element selectors here?

shiftkey (commented):
I haven't had a chance to look at this in a real browser, just curious about the benefits of tidying this up...

stylesheets/stylesheet.css
milanshresta: For better explanation: https://stackoverflow.com/questions/31317238/why-use-selector-in-combination-with-before-and-after

PR #3582, opened 2022-11-06
Stylesheet update

milanshresta:
For a better explanation: https://stackoverflow.com/questions/31317238/why-use-selector-in-combination-with-before-and-after


stylesheets/stylesheet.css
```diff
@@ -41,92 +41,15 @@ MeyerWeb Reset
...
-video {
+*,
+*::before,
+*::after {
   margin: 0;
   padding: 0;
   border: 0;
   font: inherit;
   vertical-align: baseline;
+  box-sizing: border-box;
```
shiftkey, line 53: Aside from the context of the Stackoverflow link, why do we need this rule?

shiftkey:
Closing this out due to inactivity

PR #3627, opened 2022-11-24
favicon view fixed

c9ssio:
i've swtiched the favicons to new ones that someone posted on the issues.


_includes/head.html
```diff
@@ -1,41 +1,138 @@
 <head>
-  <meta charset='utf-8' />
+  <meta charset="utf-8" />
   <meta http-equiv="X-UA-Compatible" content="chrome=1" />
-  <meta name="description"        content="Want to contribute to open source, but not sure where to start?" />
-  <meta property="og:url"         content="https://up-for-grabs.net" />
-  <meta property="og:title"       content="Up For Grabs" />
-  <meta property="og:description" content="Want to contribute to open source, but not sure where to start?" />
-  <meta property="og:image"       content="https://up-for-grabs.net/images/logo.png" />
+  <meta
```
shiftkey: There's a bunch of reformatting in here that seems unrelated to the favicon changes. I'd like to back these out so we can focus more on the right changes...

Gemfile.lock
```diff
@@ -317,6 +318,7 @@ DEPENDENCIES
   json_schemer
   rubocop
   up_for_grabs_tooling!
+  wdm (>= 0.1.0)
```
shiftkey: This dependency seems new and unrelated to the favicon work, so I'm going to back this out as I don't have any additional context on it...

shiftkey:
> i've swtiched the favicons to new ones that someone posted on the issues.

I know it's been a while since this PR was opened, but looking at the diff changes it seems like the biggest change here is a white background on the icon. This is how it looks for the default setup in Chrome on macOS:

Screenshot 2023-08-27 at 4 08 24 PM

When I switch this over to dark mode here's what I get:

Screenshot 2023-08-27 at 4 09 22 PM

Firefox in dark mode also has the same effect:

Screenshot 2023-08-27 at 4 09 58 PM

Was this what we were trying to fix @c9ssio? Without more context I'm guessing this was what we were trying to fix but it's all very hazy to me...

shiftkey (approved):

PR #3628, opened 2022-11-24
Adjust Sorting for Website "Filter by Label" field dropdown.

CxGainz:
I'm very new to JS so I apologize in advance for my unconventional code. Something I noticed when browsing the website is that the dropdown that pops up when you click the "Filter by Label" field has two label options that start with colons before the label options that begin with the letter "A" appear. An example being the ":  gift: BEGINNER" option which I believe the :  gift: was supposed to show up as an emoji. The code I included makes it so that labels that begin with non-letters will be shoved to the bottom of the dropdown so the alphabetical "a/A" labels show up first.  This was tested with Docker on Google Chrome on Windows 10.  I wanted to point out what would happen if too many project owners inputted incorrect label options, it could clutter this list.


soufainmohamd:
Cicivkvkcic

soufainmohamd:
M kcjcucic

Budgurl:
❤️

shiftkey:
@CxGainz apologies for the delay in getting to this - the changes seem fine, but it feels weird to be doing this processing each time this method runs, and I wonder if there's a different place we can do this sorting so it's done once. I'd also like to understand more about the "we render strange characters in the select inputs" case and what we might do to make this experience better, which may affect how this area works.

I need time to think about this, and sadly my time to do this sort of thing has been limited over the previous months...

P-mistry:
this.getLabels = function () {
  const labelsArray = Object.values(labelsMap);

  // Sort labels based on whether they start with a letter or not
  const sortedLabels = labelsArray.sort((a, b) => {
    const aIsLetter = /^[a-zA-Z]/.test(a.name);
    const bIsLetter = /^[a-zA-Z]/.test(b.name);

    if (aIsLetter && !bIsLetter) {
      return -1; // a comes first
    } else if (!aIsLetter && bIsLetter) {
      return 1; // b comes first
    } else {
      return a.name.localeCompare(b.name); // default alphabetical sorting
    }
  });

  return sortedLabels;
};

shiftkey:
`npm run prettier-fix` will fix this:

```
Checking formatting...
[warn] javascripts/projectsService.js
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

adityaranjan2005 (approved):

shiftkey:
Closing this out for now as I don't want to add this to the legacy site. We're working on a newer version that isn't tied to 15-year-old JS approaches, and will hopefully be an easier experience to iterate on.

PR #3646, opened 2022-12-06
Create an entry for the TX-2 Simulator project.

jamesyoungman:
I would like to add my project to Up for Grabs.

I actually have two relevant labels, "help wanted" and "good first issue".


shiftkey:
> I actually have two relevant labels, "help wanted" and "good first issue".

For the moment we only support one label here, but if you're fine with help wanted we can merge this now

shiftkey (approved):

PR #3660, opened 2022-12-12
Added Fork Count Feature Request

### LINKED ISSUE #3617: Feature Request - Fork Count
Hello,

I was using Up For Grabs, and thought it would be nice if there was a fork count along with repo name. 

Why?
It gives me an idea of how big the project is before opening the link.

I was thinking something like this, but definitely more cleaner!
image

alex-barrios:
Resolves #3617 

### Implementation
**Libraries used:**
- Python requests library
- BeautifulSoup
- pyyaml

**New files created under directory `fork_count`**
- `web_scraping_fork_count.py`
This file obtains the fork count of the repository from its GitHub repository site using requests library to get the contents of a webpage and parse using BeautifulSoup4.
Error handling: function `scraping` returns -1 if the url is invalid

- `update_fork_count.py`
This file updates the fork count of the repository’s data file using the web scraping tool from `web_scraping_fork_count.py`. This function is called in the `workflows/update-project-stats.yml` so that the fork count is updated every day. Therefore, the fork count of a new data file will created when the workflow is ran. Also, this file only updates or sets the fork count of a repository if the fork count was present on the site provided in the data file.

- `test_update_fork_count.py`
This file tests that the data files are being updated properly after using  update_fork_count.py. This test files uses a subset of 3 sample data files taken from _data/projects, which are _3sacrowd_, _30 Seconds of C++_, _42-cent_, and these files were added to a folder named _tests_sample_projects._ The file updates the fork count of these three repositories and then checks that the fork count on the data files of these repositories matches the fork count on their sites.

### Front-end
**Files edited:**

- `scripts.html` 
Added a division in HTML file under “projects” division and used `project.stats[‘fork-count’]` if `project.stats` is not NULL
Added icon for fork
The fork count was added slightly lower to the height requested on the original issue so that it is not blocked by longer project names as shown below.

	
Screen Shot 2022-12-11 at 12 25 10

### Testing
This implementation was tested using the `test_update_fork_count.py` where we tested that our implementation was working on a subset of valid repositories. We also performed tests on repositories that contained links that did not lead their GitHub repositories, so their fork count could not be determined. In these cases, the fork count statistic was not created in the data file or fork count was set to 0 if the data file contained a statistic section. These tests were added to the Continuous Integration workflow to ensure that the implementation works correctly after PR’s are developed and we checked that it runs properly in the GitHub workflow.

Furthermore, Pylint static analysis tool was used to check for errors and use proper python coding standards. Also, the Update project stats workflow was run to confirm that the implementation to update the fork count runs properly in the GitHub workflow. In addition, we tested the implementation with docker as well as with the other tests already in the codebase.


zhshakib:
You are trying to scraping the github, Have you tried github api? I think it should be much easier for you if you can do it with just a fetch request to the github api. I was trying to solve the issue, but i think you have done so much work for this feature. I think you should check the [GithubAPI](https://api.github.com/).

alex-barrios:
Hi @zhshakib! I tried using the github api but I didn't manage to get it to work because I couldn't access the token.

ritwik12:
@alex-barrios Thanks for creating this and thanks @zhshakib for your suggestion about using github API. @shiftkey will better guide you on this. Also, this can be clubbed with out daily Github Actions job to update Project Stats.

zhshakib:
I am facing issue to run the project locally with my Ubuntu 22.10 . Can i get a step by step installation guide? It will be helpful. To be noted that I am new to docker. I dont know how to run it with docker so a step by step guide will be very much appreciated. 

@ritwik12 Can you help me?

Irish78:
gre@t work

ritwik12:
> I am facing issue to run the project locally with my Ubuntu 22.10 . Can i get a step by step installation guide? It will be helpful. To be noted that I am new to docker. I dont know how to run it with docker so a step by step guide will be very much appreciated.
> 
> @ritwik12 Can you help me?

[This](https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/.github/CONTRIBUTING.md) should be able to help you for the same.

shiftkey:
@alex-barrios thanks for the submission here. As @ritwik12 mentioned, we have an existing action that knows how to find these repositories and return the open issue counts for each label. I'd recommend looking at that instead of adding a whole new infrastructure and scripting environment alongside this.

The highlights:

 - This is the code that runs daily in the repository to update each project (where it can be found on GitHub) - I think your idea about adding to `stats` here is great:

https://github.com/up-for-grabs/up-for-grabs.net/blob/8d2ba266ab61f4a04dfd2dd46f9129e0572c497b/.github/actions/update-stats/update_stats.rb#L15-L58

The logic for this is hosted in a separate repository, and the relevant places I can think of:
 
 - this is the GraphQL query that we run here, and you should be able to add a `forkCount` property to the query to get the forks

https://github.com/up-for-grabs/tooling/blob/42ae6fdf16b1272ef16800d654db12e6dec5e4a4/lib/queries/github_repository_label_active_check.rb#L88-L110

```ruby
    GitHubRepositoryLabelActiveCheck.const_set :IssueCountForLabel, client.parse(<<-'GRAPHQL')
      query($owner: String!, $name: String!, $label: String!) {
        repository(owner: $owner, name: $name) {
          label(name: $label) {
            name
            url
            issues(states: OPEN, first: 1, orderBy: {field: UPDATED_AT, direction: DESC}) {
              totalCount
              nodes {
                number
                updatedAt
              }
            }
          }
        }
        rateLimit {
          limit
          cost
          remaining
          resetAt
        }
      }
    GRAPHQL
```

 - the place where we convert the GraphQL response into a Ruby object that is returned from the method is here - adding a `fork-count` property here should also do the trick:

https://github.com/up-for-grabs/tooling/blob/42ae6fdf16b1272ef16800d654db12e6dec5e4a4/lib/queries/github_repository_label_active_check.rb#L13-L17

```
    label = repository.label
    count = label.issues.total_count
    last_updated = (label.issues.nodes[0].updated_at if count.positive?)


    { reason: 'found', name: label.name, url: label.url, count: count, last_updated: last_updated }
```


I think getting those pieces working before we update the projects here is key, so let me know if you're still interested in this addition.

shiftkey (changes requested):
Just indicating that this change isn't ready for merging, but I think there's a path forward that's simpler - it just requires some co-ordination to roll this out easily...

shiftkey:
https://github.com/up-for-grabs/up-for-grabs.net/pull/3858 has merged the changes required to get fork counts into the project data, so this is now available in the generated site if someone wants to figure out the UI parts for this.

Here's the example of how we look at the issue counts for each project:

https://github.com/up-for-grabs/up-for-grabs.net/blob/85ebaa833fab94b3a2bc6d2dce2a0912b1eb4523/_includes/scripts.html#L72-L80

PR #3668, opened 2022-12-19
adding my new project

white10146:
👋I think my project would be a great fit for up for grabs.

Let me know if there is any feedback you have about the template !

### DIFF
--- a/my-new-project.yml
+++ b/my-new-project.yml
@@ -1,0 +1,12 @@
+name:Giraffe
+desc:.NET Runtime 
+site: https://Rust.org/your-project-link-here
+tags:
+  - tags
+  - relevant
+  - to
+  - the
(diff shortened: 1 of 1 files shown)


my-new-project.yml
```diff
@@ -0,0 +1,12 @@
...
+site: https://Rust.org/your-project-link-here
+tags:
+  - tags
+  - relevant
+  - to
+  - the
+  - project
+upforgrabs:
+  name: up for grabs project itself
+  link: https://github.com/white10146/
```
shiftkey, line 12: This template doesn't seem to point to any meaningful projects

shiftkey (commented):
Not sure if the purpose of the project here is understood, so I'm going to close this out

PR #3672, opened 2022-12-22
Update default.html

starrylight90:
Head tag was missing


shiftkey:
These are handled by the `head.html` template:

https://github.com/up-for-grabs/up-for-grabs.net/blob/6bd8841bded250cc07fcaf128d04be201b89d552/_includes/head.html#L1-L41

PR #3709, opened 2023-01-02
Create elioWay.yml

timitee:
Hi. I'm Tim Bushell and the elioWay has always been learning focused. I have experience as a teacher of computing and programming at FunTech Ltd; and much of my development work has been in the education sector. I'm passionate about getting more people to program, and the elioWay is about putting application development itself into less expert hands. But the project itself is quite as simple as the library we want to create - but creating it is a great teaching opportunity. The rest I quote from our Mission Statement: 

https://elioway.gitlab.io/mission-statement.html

Intentionally break elioWay solutions into easily consumed, bite-sized libraries, which can be:

    Comprehended by new developers;
    Maintained by the inexperienced;
    Used by the skilled to accelerate productivity.

the elioWay is a chance to play with new frameworks. It is not a framework. We learn frameworks so that we can make elioWay applications, even if (or especially if) they duplicate the features of other, better, more complex solutions.

Thanks for your attention.

Tim Bushell

### DIFF
--- a/_data/projects/elioWay.yml
+++ b/_data/projects/elioWay.yml
@@ -1,12 +1,12 @@
 name: elioWay
 desc: Modular software libraries supporting an application design pattern based on Schema.org types.
-site: https://elioway.gitlab.io/
+site: https://gitlab.com/elioway
 tags:
   - javscript
   - nodejs
   - webdev
   - css
-  - "application design"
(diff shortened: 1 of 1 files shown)


shiftkey (approved):
Thanks for the submission!

PR #3741, opened 2023-01-24
Update index.html

scottykrieg:
Added boilerplate HTML, imported stylesheets.


shiftkey:
This doesn't match our required theming work that Jekyll supports, which is found in the default layout:

https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_layouts/default.html

PR #3881, opened 2023-04-10
upgrade dockerfile to use ruby 3.2 and latest bundler

shiftkey:


RiskySolutions:
You need to add a description

PR #3943, opened 2023-05-13
add my new websurfx project

neon-mmd:
Hi!, I think my project will be a good fit for up-for-grabs

Let me know if you have any feedbacks on my project!!.

### DIFF
--- a/_data/projects/websurfx.yml
+++ b/_data/projects/websurfx.yml
@@ -1,0 +1,19 @@
+---
+name: Websurfx
+desc: An open source alternative to searx which provides a modern-looking, lightning-fast, privacy-respecting, secure meta search engine written in Rust.
+site: https://github.com/neon-mmd/websurfx
+tags:
+- web-designing
+- foss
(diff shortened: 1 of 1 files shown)


_data/projects/websurfx.yml
```diff
@@ -0,0 +1,20 @@
...
+- foss
+- scraping
+- web
+- handlebars
+- css
+- actix-web
+- rust
+upforgrabs:
+  name: Up-for-Grabs
+  link: https://github.com/neon-mmd/websurfx/labels/Up-for-Grabs
```
shiftkey, line 16: This label doesn't point to any active issue - I see you're using some different labels in that repository, so feel free to change this and the `name` value above to something else if you'd prefer...

Kylemclean11: > This label doesn't point to any active issue - I see you're using some different labels in that repository, so feel free to change this and the `name` value above to something else if you'd prefer...

PR #3944, opened 2023-05-13
Add roc-java

gavv:
[roc-java](https://github.com/roc-streaming/roc-java) provides Java bindings for [Roc Toolkit](https://github.com/roc-streaming/roc-toolkit).


shiftkey (approved):
Thanks for the submission @gavv!

PR #3992, opened 2023-06-06
Add LinkFree Generator

chriskthomas:
Add LinkFree Generator, the simplest self-hosted linktree alternative.

I think that this is a good contribution because the code base is extremely simple and easy to understand for newcomers. This is about as simple as one can get while still making something useful. There are already a few users and I hope that the project can grow more with new features after the recent source code release.


shiftkey (approved):

Eminencenelly:
fixed right?

PR #4123, opened 2023-08-24
Update gitlab.yml

nickveenhof:
GitLab changed the label used to identify issues that are explicitly open to community contribution. At GitLab, all others issues are also open to contribute to, but the chance of those getting accepted by the maintainers has not been validated yet.


shiftkey (approved):
Thanks for the update @nickveenhof!

PR #4184, opened 2023-09-20
Adding Keep project

talboren:
👋 I think my project (https://www.github.com/keephq/keep) would be a great fit for Up for Grabs.

Let me know if there's any feedback you have about the template :)

Thanks!


talboren:
@shiftkey can you elaborate on what work is needed here please?

shiftkey:
@talboren please see the comment from the review automation:

> I couldn't find the label that was used in the `upforgrabs.link` value: 'good first issues'. Please update the configuration to match the right label in this repository.

There doesn't seem to be a label in your repository that matches this value

_data/projects/keep.yml
```diff
@@ -0,0 +1,12 @@
...
+desc: Open-source alert management and workflow automation platform.
+site: https://www.keephq.dev
+tags:
+  - monitoring
+  - alerting
+  - alerts
+  - workflow-automation
+  - automation
+upforgrabs:
+  name: good first issues
```
shiftkey: I think this is the problem - the `s` on the end is not actually part of the label

```suggestion
  name: good first issue
```

_data/projects/keep.yml
```diff
@@ -0,0 +1,12 @@
...
+  link: https://github.com/keephq/keep/issues?q=label%3A%22good%20first%20issue%22
```
shiftkey, line 12: :art: a simpler URL that's equivalent to the current value

```suggestion
  link: https://github.com/keephq/keep/labels/good%20first%20issue
```

talboren: Gotcha, thanks!

talboren:
@shiftkey i think it’s ok now :)

shiftkey (approved):

PR #4251, opened 2023-10-06
adding trivy-resource project

hriprsd:
I think i would like to add my OSS project to up-for-grabs. Thanks!


hriprsd:
@shiftkey could you please help review my pr? Thanks in advance

shiftkey:
@hriprsd this is the comment that needs to be addressed

> Required fields are missing from file: site. Please check the example on the README and add these values.

This means a value is expected but not missing. It should look like this:

```
site: [url]
```

_data/projects/trivy-resource.yml
```diff
@@ -0,0 +1,11 @@
...
+desc: Trivy resource for concourse - Vulnerability scanning of container base image
+site: https://github.com/Comcast/trivy-resource/
+tags:
+  - trivy-resource
+  - trivy
+  - concourse-resource
+  - vulnscan
+upforgrabs:
+  name: up for grabs
+  link: https://github.com/comcast/trivy-resource/labels/up%20for%20grabs
```
shiftkey: There's no label listed here which matches this - https://github.com/comcast/trivy-resource/labels

hriprsd: My bad. Added it in. Thanks!

hriprsd: Fixed all the review comments. Please lemme know if i need to fix anything more. Thanks in advance 😄

shiftkey: Almost there

```suggestion
  link: https://github.com/Comcast/trivy-resource/labels/up%20for%20grabs
```

Issue #3605, opened 2022-11-14
Improving Card Tags


Yasir761:
![Issue - 1](https://user-images.githubusercontent.com/74600745/201686217-24c96e0d-0c96-40d5-8ee7-92fd21f82afd.PNG)
Cards Tags should be at the bottom. This means whether the content of the card is  small or big, card tags are positioned at the bottom.

ramialkarshi:
مرحبا ياسر

aruadecarvalho:
Hi! Can I work on this issue?

SaintQ969:
Hello, I would love to work on this if you don't mind?

sh471:
Sure, it sounds great.

*Best Regards,*

*Sherif Hikal*

SaintQ969:
Thank you so much for the opportunity.
Is there a way I can mail you directly?

Yasir761:
Since I'm not a Maintainer of this repo so, I cannot assign this to anyone. I also want to work on this issue that's why I raised it
@aruadecarvalho

Yasir761:
@SaintQ969 @aruadecarvalho I cannot assign it to anyone of you because I am not an owner of this repo.

Momo1kz:
é&²zkk
@ebekker

sh471:
> Thank you so much for the opportunity. Is there a way I can mail you directly?

Absolutely, thank you for reaching out.

Momo1kz:
@camilosampedro fg

Memomer:
Can i work on this issue?

Upendra672:
I would love to work on this issue if you don't mind

sh471:
> Can i work on this issue?

luuu-xu:
Hi, is someone still working on this issue? I'd love to work on it if nobody is.

RachitSahu26:
don't ask that can i do this. if you can do then do

> Can i work on this issue?

aditya-desai-19:
I would like to work on this. Please assign this issue to me.

sh471:
Absolutely, go ahead.


*Best Regards,*

*Sherif Hikal*

shiftkey:
Resolved via #4283

PR #4283, opened 2023-10-21
fixed the card-tags at the bottom

Aditya1979314:
issue - #3605 

Fixed the tags at the end of the data-card .

upForGrabs


_includes/scripts.html
```diff
@@ -61,6 +61,7 @@
     <% _.each(projects, function(project) { %>
     <% count= count+1%>
     <div class="data-box" class="<% if ( project.stats ) { %>counted<% } %>">
+      <div>
       <div>
```
shiftkey: :art: indenting this element (and related child nodes) would be nice to do as a tidy-up

shiftkey (commented):
One additional suggestion, otherwise changes look fine in the major browsers

_includes/scripts.html
```diff
@@ -58,54 +58,62 @@
...
+                    <% } %>
+                </div>
+              </div>
+            </div>
+            <div class="details-tags">
+              <% if (project.tags) { %>
+                <p class="tags">
+                  <% _.each(project.tags, function(tag, i) { %>
+                    <a href="#/filters?tags=<%-encodeURIComponent(tag)%>" tabindex="-1"><%-tag%></a><%- i
+                      !=project.tags.length-1 ? "," : "" %>
```
shiftkey: :art: let's not break the line mid-expression

```suggestion
                    <a href="#/filters?tags=<%-encodeURIComponent(tag)%>" tabindex="-1"><%-tag%></a><%- i !=project.tags.length-1 ? "," : "" %>
```

_includes/scripts.html
```diff
@@ -114,12 +122,14 @@
...
 <!-- Google Analytics -->
 <script>
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+  (function (i, s, o, g, r, a, m) {
+    i['GoogleAnalyticsObject'] = r; i[r] = i[r] || function () {
+      (i[r].q = i[r].q || []).push(arguments)
+    }, i[r].l = 1 * new Date(); a = s.createElement(o),
```
shiftkey, line 129: This is an unnecessary change, but given I can't recall where this GA portal is hiding I think I'll just rip this out as a follow-up change.

_includes/scripts.html
```diff
@@ -58,54 +58,61 @@
...
+                </div>
+              </div>
+              <div>
+                <div class="details stats">
+                  <p class="label">
+                    <a href="<%- project.upforgrabs.link %>" title="View open issues for <%-project.name %>"
+                      target="_blank" tabindex="-1">
+                      <%-project.upforgrabs.name %>
+                        <% if (project.stats && project.stats['last-updated']) { const now=new Date(); let lastUpdated=new
+                          Date(project.stats['last-updated']); let since=relativeTime(now, lastUpdated); %>
```
shiftkey: This compression of multiple lines also feels weird - can we bring back the previous formatting and make this code easier to follow?

_includes/scripts.html
Aditya1979314: So should I leave this change as it is , or if you want I can change it to the previous code as well .

shiftkey: Let's leave this alone for now

_includes/scripts.html
Aditya1979314: Is it good now ?

shiftkey: No - see how the previous style broke the code over multiple lines?

```js
 <% if (project.stats && project.stats['last-updated']) {
                const now = new Date();
                let lastUpdated = new Date(project.stats['last-updated']);
                let since = relativeTime(now, lastUpdated);
              %>
```

That context is lost here

Aditya1979314: Sorry for the fault , but I have fixed this issue now.

shiftkey (approved):

PR #4345, opened 2023-11-23
fixx #3628s

P-mistry:
this.getLabels = function () {
  const labelsArray = Object.values(labelsMap);

  // Sort labels based on whether they start with a letter or not
  const sortedLabels = labelsArray.sort((a, b) => {
    const aIsLetter = /^[a-zA-Z]/.test(a.name);
    const bIsLetter = /^[a-zA-Z]/.test(b.name);

    if (aIsLetter && !bIsLetter) {
      return -1; // a comes first
    } else if (!aIsLetter && bIsLetter) {
      return 1; // b comes first
    } else {
      return a.name.localeCompare(b.name); // default alphabetical sorting
    }
  });

  return sortedLabels;
};

"modify changes"

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ If you know of or own a project that should be listed on Up for Grabs,
 If you would like to get involved with the project, please read the
 [CONTRIBUTING.md](.github/CONTRIBUTING.md) file as it contains:
 
- - instructions about getting your environment setup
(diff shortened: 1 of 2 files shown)


P-mistry:
this.getLabels = function () {
  const labelsArray = Object.values(labelsMap);

PR #4430, opened 2024-01-05
Added Gen1Py

9x14S:
I believe this one will be fun for those who want to start reverse engineering/game hacking. This will give them the opportunity to try reversing the way the games encode their data, and also might be interesting for those who would want to make a GUI for it.


9x14S:
Okay, I give up. What's wrong with my label?

PR #4437, opened 2024-01-10
Adding a new project (scanner)

davide710:
Hi, I'd like to add my project to Up for Grabs.
I think it would be a great opportunity for new contributors because of the small size of the codebase and its simplicity.
It's a useful project that could use some improvements and I'd be glad to collaborate with other passionate people like me.
Please let me know if something is wrong with my PR.
Thanks in advance,
Davide

### DIFF
--- a/_data/projects/scanner.yml
+++ b/_data/projects/scanner.yml
@@ -1,0 +1,11 @@
+name: scanner
+desc: Transform a photo of an A4 document to a pdf
+site: https://github.com/davide710/scanner
+tags:
+  - python
+  - opencv
+  - computer-vision
+  - scanner
+upforgrabs:
+  name: up-for-grabs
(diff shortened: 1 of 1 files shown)


davide710:
I can't tell if my pull request is fine and I just have to wait or if there's something wrong, and if so what can I do about "javascript" and "site" checks? I followed https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/docs/list-a-project.md and I don't know what I could do differently.

shiftkey (approved):

shiftkey:
> I can't tell if my pull request is fine and I just have to wait or if there's something wrong, and if so what can I do about "javascript" and "site" checks? I

These checks are not performed by default for new contributors due to abuse in the past, but otherwise this is fine to proceed...

Issue #4303, opened 2023-10-28
Add tooltip for fork numbers


koppor:
When browsing the up-for-grabs web site, one sees a pull request icon and a number after it. It is not clear, what this number means.

Propposal: Add a tool tip on hover (and maybe keep the mouse pointer as pointer - and not change it to a text cursor)

![image](https://github.com/up-for-grabs/up-for-grabs.net/assets/1366654/f4e09207-2449-4390-85bc-e804b10b9956)

PushpRatan:
I have fixed this issue, this is my first contribution. Please guide me if something is wrong. Thank you.

llr3v0ll:
I've opened a pull request for this issue

PR #4465, opened 2024-01-28
added-fork-tooltip-on-hover-issue#4303

llr3v0ll:
The tooltip will appear on top of the items upon hovering; I think it looks the best on top even though it might be over some titles. If you don't like it on top you can tell in which direction to put it and I can change it.

https://github.com/up-for-grabs/up-for-grabs.net/assets/130805782/72744694-aadc-4974-b385-154622aa2fe5

image
image


koppor:
Fixes https://github.com/up-for-grabs/up-for-grabs.net/issues/4303

PR #4535, opened 2024-03-17
adding my project - Create taskorial.yml

nayasinghania:
I think my project would be a good fit for up for grabs. Please let me know if there are any other changes that need to be done before this can go on up for grabs


raspberri05:
I fixed the issue that the bot is talking about here but not sure why the issue is still here and how to fix it. Hoping for some  help

shiftkey (approved):

PR #4565, opened 2024-04-05
Added Glasskube to Up For Grabs

jakepage91:
Hey there! If you could find some time to consider including Glasskube to the list of available projects I would hugely appreciate it. On our side we are going to do our best to keep a healthy and varied array of `good first issues` for the contributors to choose from.


_data/projects/glasskube.yml
```diff
@@ -0,0 +1,18 @@
...
+tags:
+  - kubernetes
+  - kubernetes-api
+  - cli
+  - go
+  - typescript
+  - docusaurus
+  - makefile
+  - mdx
+  - cli
```
shiftkey:
```suggestion
```

PR #4697, opened 2024-07-10
PR to fix failed automatic update of eslint to v9.6.0 #4688

anasKhafaga:


package.json
```diff
@@ -19,6 +19,8 @@
     "npm": ">=7"
   },
   "dependencies": {
+    "@eslint/eslintrc": "^3.1.0",
+    "@eslint/js": "^9.6.0",
```
shiftkey: 🤔 Should these also be `devDependencies`?

eslint.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+        "func-names": "off",
+        "one-var": "off",
+        eqeqeq: "off",
+        "no-param-reassign": "off",
+        "no-underscore-dangle": "off",
+        "no-useless-escape": "off",
+        "no-restricted-globals": "off",
+        "no-use-before-define": "off",
+        "global-require": "off",
+        "import/no-amd": "off",
```
shiftkey: 🤔 This rule is from `eslint-plugin-import` so I'd expect some sort of error if we've had to remove the plugin...

package.json
anasKhafaga: That's my bad, I pushed more focus to migration of `eslint` configuration. It's fixed now.

eslint.config.mjs
anasKhafaga: Actually there was no thrown error so I haven't noticed it. But you are right, there's no need for it. Removed!

Thanks for your review @shiftkey

shiftkey (approved):

Issue #4489, opened 2024-02-12
UI Enhancement


Mayankkatheriya:
![image](https://github.com/up-for-grabs/up-for-grabs.net/assets/128832286/0e247423-1683-46b6-80ad-3138356272af)

it becomes more attractive if we add some hove effect on these div
like on hover its scale becomes larger and box shadow increase to give 3D like structure

Kamainkokamain:
Bouruoree

Mayankkatheriya:
What?

mukhammadjcn:
Hi, there how can I send my version to you

shiftkey:
Resolved via https://github.com/up-for-grabs/up-for-grabs.net/pull/4513

PR #4712, opened 2024-07-16
Added hover 3d effects

dayyan1341:
Added a scale change of 0.05 on hover as well as moved the box shadow to when hovered

### DIFF
--- a/stylesheets/stylesheet.css
+++ b/stylesheets/stylesheet.css
@@ -778,14 +778,21 @@ form {
 
 .data-box {
   background-color: var(--databox-bg);
-  box-shadow: 0px 5px 10px var(--box-shadow-color);
   padding: 10px;
   border: 1px solid var(--container-border);
   border-radius: 10px;
   display: flex;
   flex-direction: column;
   justify-content: space-between;
+  transition-property: box-shadow, scale;
+  transition-duration: 300ms, 300ms;
+  transition-timing-function: ease-in-out, ease;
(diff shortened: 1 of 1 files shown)


dayyan1341:
closes #4489 

Normal :- 
image


Hovered :- 
image

shiftkey:
This seems to cause the "last updated" options to stack.

They currently look like this:

Screenshot 2024-07-16 at 8 39 21 AM

But with this PR they now look like this:

Screenshot 2024-07-16 at 8 39 15 AM

I also don't seem to see the same styling effect on the cards themselves:

Screenshot 2024-07-16 at 8 40 19 AM

They do the zoom effect on hover, but there's lots of other regressions here that mean I can't accept this in it's current form.

You can test the changes out by pushing more commits to this branch, which are then published here:

https://deploy-preview-4712--up-for-grabs-test-bench.netlify.app/?pr=4712#/

stylesheets/stylesheet.css
```diff
@@ -778,15 +778,22 @@ form {
 
 .data-box {
   background-color: var(--databox-bg);
-  box-shadow: 0px 5px 10px var(--box-shadow-color);
```
shiftkey, line 781: This box seems to be used in many other places, and could be the root cause of all these regressions.

dayyan1341:
image

https://deploy-preview-4712--up-for-grabs-test-bench.netlify.app/?pr=4712#/

PR #4818, opened 2024-09-25
Replace `CASE` name by `manifest`

SebConejo:
The project called `CASE`was listed in up-for-grabs`. But we changed the name of the product. It's now called Manifest. 

This PR remove the CASE.yml file and add a manifest.yml file to keep an updated list.

- The project has evolved but is still similar. We target the same target and we are Open source too. If you need more information before accepting the PR let me know.


shiftkey (approved):

shiftkey:
There's a merge conflict here that I can't resolve without manually cloning the repository and rewriting the branch, which I don't have the bandwidth for currently...

shiftkey (approved):

PR #4865, opened 2024-10-15
Add Qtcord

mak448a:
Hi, could you add my project to the up for grabs list?
Qtcord is a lightweight Discord client written in Python, Requests, and Qt.


mak448a:
Better?

PR #4902, opened 2024-11-08
Move theme storing to local storage

phamduylong:
So I realized that theme was being stored in **sessionStorage** and it was not persisted after each visit to the site. This PR does 2 things:

- Move the stored theme to **localStorage**.
- If no theme was stored in the **localStorage**, it [checks user's theme preferences](https://stackoverflow.com/a/57795495/14126819).

### DIFF
--- a/javascripts/dark-mode.js
+++ b/javascripts/dark-mode.js
@@ -14,7 +14,8 @@ define([], () => {
     lightModeEnabled = storedValue === 'light';
   } else {
     // no stored theme - let's check user's preference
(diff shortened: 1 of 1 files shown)


phamduylong:
@shiftkey Sorry for not discussing this beforehand, but what do you think about this change? It's a bit annoying that the theme keep switching back to light theme when I open the page tho I set it to dark on the last visit.

shiftkey (approved):

PR #4925, opened 2024-11-23
Page navigation

### LINKED ISSUE #4923: Navbar or Scrollbar
We can add a navbar on upforgrabs.net that switches between "I want to get involved", "projects" and "I want to maintain a project". User's don't have to manually scroll the entire page to get relevant information 

Another option is that instead of switching between web-pages on a navbar, we could add a scroll bar. Users can click on different sections on the scroll bar, and the website would automatically scroll them to their desired sections aesthetically.

nobodys-ghost:
This pull request adds a navigation bar to index.html to allow users to jump between sections of the site without having to scroll past the lengthy projects section.

Closes #4923


shiftkey:
@nobodys-ghost the nav bar looks really big here, not sure of what it _should_ look like but it doesn't feel great in my opinion:

Screenshot 2024-12-31 at 5 06 21 PM

I also added notes here about the utility of this navbar, which I think are relevant here https://github.com/up-for-grabs/up-for-grabs.net/pull/4976#issuecomment-2565838135

PR #5001, opened 2024-12-30
Updated Launchit

iddu0:
Updated the following:
- Fork count (now 1)
- Issue count (now 3)


shiftkey:
@iddu0 these numbers are auto-updated in the background every couple of days - no need to submit manual PRs, I'll kick off a new job to regenerate and commit for these new projects

shiftkey:
Commit just landed with the new values:

https://github.com/up-for-grabs/up-for-grabs.net/blob/82467cd8975a4eff1386da953927f62e7b9f61b9/_data/projects/Launchit.yml#L11-L14

PR #5027, opened 2025-01-12
Create uregister.yml

falseywinchnet:
so, i'm considering writing such a tool, but I know i wouldn't maintain it and I would only get marginal utility out of it. I think a lot of people would probably use such a library for installers and service management panels. I am still considering contributing code to such a program, but I don't want to own such a library.

is there a better place to propose such a project?


shiftkey:
> is there a better place to propose such a project?

Probably not. We don't handle the ownership side of these projects.

PR #5114, opened 2025-03-16
popular tags button functionality added

TomSamwan:
Hi there! This is my first contribution to an open-source project, so I hope I'm submitting everything correctly.

I have amended the main.js code for the popular tags buttons so that when you press them, they are added to the url query string and also added to the tag filter field.

I've also added my details into the contributions file too.

please let me know if there's anything I've missed/needs changing or if you find my code doesn't work correctly.


javascripts/main.js
```diff
@@ -153,27 +153,32 @@ define([
...
-              getFilterUrl(),
-              'tags',
-              encodeURIComponent(selTags)
-            );
-          }
+      $(elem).on('click', function (e) {
+        const tags = prepareForHTML(getParameterByName('tags')) || [];
+        selectedTag = (e.currentTarget.innerText).split("\n")[0] || '';
+        if(!tags.includes(selectedTag)) {
+        var component = $(`
```
shiftkey:
```suggestion
        const component = $(`
```

shiftkey (approved):
Taking this in to fix the bug, but it now looks like all our other tag links are no longer additive (and they will overwrite the selected tags with the single tag).

Issue #4334, opened 2023-11-13
Add a github icon or button to go to project github repository?


piyushbarve98:
**Description**
Currently, on the UpForGrad project page, clicking on a project name most of the times it directly opens the live website and in some cases it opens github repo. It would be beneficial to provide easy access to the GitHub repository for users who want to explore the codebase and contribute.

**Proposed Solution**

I suggest adding a GitHub icon or a link next to the project name, which, when clicked, takes the user to the GitHub repository for that project.

Assign this issue to me or I'll raise a PR directly.
Also I am open to suggestions regarding the issue.

Mano3009:
Hi can you assign this task to me , I am working on it

shiftkey:
> Currently, on the UpForGrad project page, clicking on a project name most of the times it directly opens the live website and in some cases it opens github repo.

This is a conscious decision to favour whichever `site` value the project has submitted. We wanted to distinguish this from the issues available as a way to point folks to the "home page" for the project, but there's a good number of projects which skip the "home page" and use the GitHub repository instead (which is fine).

I'm skeptical of the changes here without more detail about how the proposed changes would work, so feel free to flesh this out further if you have specific ideas here.

Chand5120:
Cn you assign this issue to me, i can work on it.

Chandi977:
Can you assign this issue for me.

PR #5179, opened 2025-04-30
Fix Issue #4334: Add GitHub icon linking to git repository

foxzhang1224:
- Related to #4334

🔧 What has been changed?
This PR adds a GitHub icon next to each project name on the homepage that links to its github repository page.
Clicking the GitHub icon takes the user directly to the root GitHub repository of the project (e.g., https://github.com/user/repo).

🎯 Why This Matters
Many projects listed on up-for-grabs.net currently link to a GitHub label like:
[https://github.com/user/repo/issues/labels/good-first-issue](url)

Therefore, it there could be a way to directly link the user to the GitHub repository, a quicker navigation could be offered to explore the full project on a technical term, especially to those potential contributor.

🛠️ How It Works
 1) Extracted the repository root by splitting the project.upforgrabs.link and joining the first 5 segments:
`project.upforgrabs.link.split('/').slice(0, 5).join('/')`
Since we couldn't directly get the url of many projects, parsing the link to its labelled issue could lead us to the main repo page.

2) Conditionally rendered the icon only if project.upforgrabs.link is defined

3) Added the icon using a github SVG from Simple Icons cdn

`GitHub`

✅ Preview
Screenshot 2025-04-30 at 5 53 09 PM


shiftkey:
I don't think this is the correct fix, as this seems to be applied to all projects (including non-GitHub projects).

Let's go back to #4334 to talk about this further, as I don't think this is the right fix for the problem.

PR #5217, opened 2025-05-27
Clicking on "Popular tags" now applies the related filter (#5216)

### LINKED ISSUE #5216: Popular tags appear clickable but are not functional
When hovering over the "Popular tags" icons under the filters, the UI changes (border turns blue), which gives the impression that they are clickable buttons. As a new user, I instinctively tried to click on them expecting them to apply a filter. I think it would improve the user experience if clicking a popular tag applied the corresponding filter, especially because this section is visually grouped with the filters above.

DooD73:
## What this PR does

This implements the feature proposed in #5216. Clicking on any "Popular tag" under the filters now applies that tag as a filter, aligning with user expectations set by the hover effect and visual grouping with the filter section.

## Summary of changes

- Added href to handle filter by tags

## Preview

https://github.com/user-attachments/assets/b6a2097a-c5de-4ec8-8943-6da99ac79aa9

Closes #5216


shiftkey:

PR #5252, opened 2025-06-17
Create shiori.yml

Brianali-codes:
As the sole developer, i lack some of the optimization techniques and the ability to complete some of the features I would like to resolve the performance problem with the app i already have the pc client that was expected to pair with the mobile app to enable the cross sharing support feature, but currently i lack the required resources, therefore any form of help will be much appreciated, i also would like it if someone splits the apk to different abis or architectures in order to reduce the apk size.


Brianali-codes (commented):
update link

shiftkey (approved):

PR #5308, opened 2025-07-28
Add kaju2116 project to up-for-grabs

kaju2116:


kaju2116:
this is my first contribution!

_data/projects/kaju2116.json
shiftkey, line 1: This needs to be a YAML file

PR #5370, opened 2025-09-27
Add go-vend-event.yml for festival vendor app

AnnaVi11arrea1:
I would like to add my app to up for grabs, :)


go-vend-event.yml
shiftkey, line 1: This needs to be added under `_data/projects/` to be part of the site data

PR #1121, opened 2018-12-17
Fixed popular tags not working

juliankrieger:
I fixed the popular tags not working by adding an on("click") eventHandler that changes the location.href according to the tag clicked. I'd be glad if it could be reviewed by either @ritwik12 or @Zarad1993 

Fixes #1116 in a clean and readable way.
~However, currently only one popular tag works at a time. I'll look into this this evening~

However, I'm not sure if it's enough to generate the correct link and change it accordingly, or if I should push the new tag to the projectServices "tags" array, too.

PR #5380, opened 2025-10-01
Fix click on popular tags

### LINKED ISSUE #5216: Popular tags appear clickable but are not functional

mkazin:
Fixes #5216

This PR restores the functionality of clicking on Popular Tags by correcting a bug in tag parsing. It also enables a second click to remove the tag from the filter list.

The existing code uses code to handle the project count element nested within the anchor containing the tag. This new code is more conservative by extracting the tag from the first TEXT_NODE, to avoid making assumptions about other child elements which was necessary by using innerText.

Unlike #5217, this PR does not conflict with the existing event listener or the override already-filtered tags introduced by #1121 (and later fixed in #1179). If the other approach is preferred, it should be amended to remove the popular handling from main.js .

### DIFF
--- a/javascripts/main.js
+++ b/javascripts/main.js
@@ -156,13 +156,24 @@ define([
     projectsPanel.find('ul.popular-tags li a').each((i, elem) => {
       $(elem).on('click', function () {
         selTags = $('.tags-filter').val() || [];
-        selectedTag = preparePopTagName($(this).text() || '');
+        selectedTag = preparePopTagName(
+          Array.from(this.childNodes)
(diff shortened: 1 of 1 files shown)


shiftkey:

Issue #5369, opened 2025-09-24
[BUG] Popular tags are unresponsive


Mayank85Y:
<h3>Description</h3>
The tags listed under Popular Tags are not interactive/responsive when clicked. This prevents expected navigation or filtering behavior.

<h3>Steps to Reproduce:</h3>

- Go to https://up-for-grabs.net/
- Locate the Popular tags section
- Click on any tag availabl in Popular tag section

<h3>Context</h3>

https://github.com/user-attachments/assets/26d322ca-716a-4bc0-898d-b391c94404d0

<h3>Expected Behaviour</h3>
When someone click on the Popular tag. It should apply that filter and query the data.

panagiotisbellias:
I would like to work on it, can you assign it to me?

resul-elezi:
@Mayank85Y check issue #4917

RAGHUKHAJURIA:
i would like to work on this issue

Huzaif-nabi:
Is this issue still open? Can i work on it

ssagar1999:
hi, i would like to work on this issue. if it is okay , please asign it to me, i will aim to complete it within the next two days

ivanho-git:
the issues seems solvable and interesting can you assign it to me??

shiftkey:
I've merged https://github.com/up-for-grabs/up-for-grabs.net/pull/5114 to address this issue

PR #5382, opened 2025-10-03
Fixed issue: Popular tags are not interactive

ameyshigam:
Issue solved #5369 
BUG Popular Tags are unresponsive.
  1. The problem was in the click event handler for the popular tags in main.js . The original code had several issues:
   
   - It was using undeclared variables ( selTags , selectedTag , tagID )
   - The preparePopTagName function was only extracting the first word of the tag name
   - It wasn't preventing the default link behavior
2. My fix includes:
   
   - Adding e.preventDefault() to prevent the default link behavior
   - Using const to properly declare variables
   - Improving the tag text extraction by splitting on the parenthesis and trimming
   - This ensures the full tag name is captured correctly, not just the first word
   - 
The tags should now be interactive and properly filter the projects when clicked. This matches the expected behavior where clicking on a popular tag should filter the projects list to show only projects with that tag.


shiftkey:

Issue #5515, opened 2026-01-25
Project render order is not based on alphabetical order


shiftkey:
In the current site we introduce a random order for initially rendering the projects to avoid alphabetical order and favouring showing projects that start with A. This is persisted to client-side storage so that it works consistently across reloads.

To leverage Astro better, we should randomize the order each time the site is published and avoid the client-side dependency. But I haven't thought through how the sort order should work here - open to ideas.

PR #5584, opened 2026-03-03
fix: stop persisting project order in sessionStorage

nthbotast:
## Summary
- remove `sessionStorage` reads/writes from `orderAllProjects`
- always derive ordering from the provided `computeOrder` function
- simplify/refresh unit tests to match the new behavior

This addresses #5515 by removing client-side storage as the source of project render order.

## Testing
- npm test -- project-ordering-spec


shiftkey:
#5515 is for the new beta site. Let's not remove this functionality for the existing site.

I've updated #5515 to add some more thoughts here related to sort order of projects in the new site.

PR #5585, opened 2026-03-04
fix: quote multi-word GitHub labels in project issue links

### LINKED ISSUE #5239: Labels with spaces in them are nonfunctional
To recreate:
Pick a project where the label has spaces in it. For example, this project "Cutter" https://up-for-grabs.net/#/filters?names=225 has label "good first issue". Click "Good first issue" to be taken to the github issues page. It takes you to https://github.com/rizinorg/cutter/labels/good%20first%20issue instead of https://github.com/rizinorg/cutter/labels/%22good%20first%20issue%22 (`/"good first issue"`)
The critical difference: The first searches for issues matching `is:issue state:open label:good first issue`, where first issue are keywords matched in text and the label searched for is just "good" (which for most repos is invalid). The second URL will correctly search issues for `label:"good first issue"`.

I was going to try and dig into the code and submit a PR, but it looks like all URLs are configuration based. So it will be up to the maintainer how to address this. Run a script to update all the configuration files? Just add a function to change the variable values on the site itself to wrap labels in quotes?

Image

Image

nthbotast:
## Summary
- normalize GitHub label links before rendering in project cards
- wrap multi-word labels in quotes (URL-encoded) so GitHub filters the exact label
- leave non-GitHub or already-quoted label links unchanged

Fixes #5239

## Testing
- npm run lint-new -- src/components/search/ProjectEntry.vue

### DIFF
--- a/src/components/search/ProjectEntry.vue
+++ b/src/components/search/ProjectEntry.vue
@@ -11,6 +11,35 @@ const { project } = props;
 const stats = project.stats;
 
 const linkTitle = `View open issues for ${project.name}`;
+
+const getIssueLabelLink = (link: string): string => {
+  try {
+    const url = new URL(link);
+
+    if (url.hostname !== 'github.com') {
+      return link;
+    }
+
+    const match = url.pathname.match(/^(\/[^/]+\/[^/]+\/labels\/)([^/]+)$/);
+    if (!match) {
+      return link;
+    }
+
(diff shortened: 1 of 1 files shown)


shiftkey:
I've closed out the upstream bug as I believe that the label behaviour here is working correctly. It'd be great to add some component tests so we can automatically validate this fix addresses a bug with the new site.

Issue #4917, opened 2024-11-18
Popular Tag count not being updated


Hasteerp:
_Current Behaviour_ : 
Currently, popular tag count are not being updated based on the filters selected by the user. 

_Image_ : 
![ScreenRecording2024-11-18at5 13 05PM-ezgif com-video-to-gif-converter](https://github.com/user-attachments/assets/2906d073-4f13-4ae4-be72-e72588880dd0)

_Expected Behaviour_ :
Popular tag count updates dynamically on change of filters.

I can take it up and make a PR with the necessary changes!

soumyamakkar:
Can I be assigned this issue?

7umair-ahmed7:
Can I work on this issue?

TomSamwan:
I've created a pull request which fixes this.
Can I get my code reviewed please?

Huzaif-nabi:
Can i work o it?
I can take it up and make a PR with the necessary changes!

shiftkey:
Can someone share visual details of what they're seeing?

I have these PRs but it'd be great to understand the problem better first:

 - https://github.com/up-for-grabs/up-for-grabs.net/pull/5340
 - https://github.com/up-for-grabs/up-for-grabs.net/pull/5587

Reshi12:
Team Geekmaxxed has deprioritized this issue.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
Team Geekmaxxed is working on this issue. A pull request will be submitted shortly.

Reshi12:
[...]

PR #5587, opened 2026-03-04
fix: update popular tag counts when filters change

nthbotast:
## Summary\n- compute popular tags from the currently filtered project list instead of the full dataset\n- keep the same top-N behavior while recalculating frequencies for visible results\n- add unit tests for filtered popular-tags behavior and no-match edge case\n\n## Testing\n- npm test -- --runInBand tests/spec/ProjectsServiceSpec.js\n\nCloses #4917


shiftkey:
This PR currently has conflicts against multiple files so I'm going to close it out as stale. Please review what is currently deployed and submit a new PR if there are changes that you still wish to incorporate into the current site.

PR #5618, opened 2026-03-11
Sort projects by recently updated instead of alphabetically

stevencl:
## Summary

Changes the default sort order for projects from alphabetical by name to **most recently updated first** (descending by `last-updated` date).

## Changes

- **`src/components/data/projects.ts`** — Build-time sort now orders `RawProjects` by `stats['last-updated']` descending.
- **`src/components/data/search.ts`** — Search results sort also orders by `stats.lastUpdated` descending, so filtering/searching preserves the "recently updated" order.

Projects without a `last-updated` date fall back to alphabetical ordering so nothing is lost.

## Motivation

The alphabetical sort made it hard to discover newly added or recently active projects. Sorting by most recently updated surfaces the most active projects at the top, which is more useful for contributors looking for something to work on.

### DIFF
--- a/src/components/data/projects.ts
+++ b/src/components/data/projects.ts
@@ -23,7 +23,14 @@ export const RawProjects: Array<Project> = projects
       ...project.data,
     };
   })
-  .sort((left, right) => left.name.localeCompare(right.name));
+  .sort((left, right) => {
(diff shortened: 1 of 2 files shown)


stevencl:
Sorry, added by mistake.

PR #5624, opened 2026-03-15
fix(search): include tag values in text search results

### LINKED ISSUE #5514: Search for projects based on tag

nthbotast:
## Summary
- include project tag values in SearchProjects filtering
- keep existing name/description checks unchanged
- allow tag text queries (e.g. language/platform tags) to return matching projects

Fixes #5514

## Validation
- npm run -s lint-new
- git diff --check

### DIFF
--- a/src/components/data/search.ts
+++ b/src/components/data/search.ts
@@ -115,5 +115,9 @@ export function filterProject(
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@nthbotast I've added support for testing this component - are you able to add some tests to `src/components/data/search.test.ts` to cover this case, so we can catch regressions in the future?

Issue #3495, opened 2022-10-14
[BUG] Filter by name select box


Anuran12:
###ISSUE:

Filter by name box is getting outside from the page in mobile view.. looks like its not responsive...

![WhatsApp Image 2022-10-14 at 12 27 46 PM](https://user-images.githubusercontent.com/104062485/195783066-19a64c1c-eff1-4f1c-b1a5-a79429d342a8.jpeg)

I would like to fix it.. please assign this to me.

Aee-hrishya:
@Anuran12 Have you started working on this?

Kennyev79:
Assign me

Anuran12:
> @Anuran12 Have you started working on this? 

You didn't assign so i didn't work. But it's a very easy tusk. It will take a few minutes

lilitha-mdlalana:
> 

Have you done it yet?

ABSTRAC7ION:
if you assign me i will start working on it now, it will take any time

yuvi3005:
fast fix, chk other repo

Yasir761:
Please let me know if this issue is not fixed yet because if it is not then I can assure you that I can fix this.

ramialkarshi:
3495

sh471:
yes sir, we will do thank you.



*Best Regards,*

*Sherif Hikal*

enricozz:
I think I was able to fix it. I added a width property of 100vh inside of the media query. On chrome inspector, all devices work.

dbillion:
please assign me to it

shiftkey:
I'm not seeing this issue with emulated Chrome or Safari on my local machine - is this still observable?

<img width="536" height="684" alt="Image" src="https://github.com/user-attachments/assets/807e3436-06f6-4d30-b1e2-8f2073de98de" />
<img width="731" height="579" alt="Image" src="https://github.com/user-attachments/assets/9a6c8240-b68f-4bc3-833a-954f75939111" />

GowsiSM:
@shiftkey Thanks for checking!
From the screenshots it looks like the filters are rendering as native select elements locally, while the deployed site uses the chosen.js styled dropdowns. 

The change constrains the select/container width so the dropdown cannot extend beyond the viewport on smaller devices.

PR #5640, opened 2026-03-16
fix(css): keep filter selects within viewport on mobile

nthbotast:
## Summary\n- make Chosen multi-select filter containers use full available width\n- ensure the filter choice box uses border-box sizing and does not overflow\n- keep desktop behavior unchanged while fixing mobile viewport overflow\n\nFixes #3495\n\n## Validation\n-


shiftkey:
I've merged in #5641 which was competing with this change, but refinements are open if you see anything weird in the current site

PR #5641, opened 2026-03-16
Fix mobile overflow for filter select boxes (#3495)

### LINKED ISSUE #3495: [BUG] Filter by name select box

GowsiSM:
Fixes #3495

This change fixes the issue where the filter select box can go outside the page on smaller screens.

What I changed:
- Added `box-sizing: border-box` to the select element
- Set `width: 100%` along with `max-width: 100%`
- Prevented overflow from the filter panel
- Made the filter label appear above the dropdown on smaller screens

These changes help keep the dropdown inside its container so it does not cause horizontal scrolling on narrow mobile widths. The layout on larger screens remains unchanged.


shiftkey:
Not sure what's changed here but this seems to break the select inputs when I test Safari at mobile resolution.

Repro steps:

   - open the staging site: https://deploy-preview-5641--up-for-grabs-test-bench.netlify.app/#/
   - click into one of the select menus

Expected behaviour:

 - options are visible - this seems to work fine on the deployed site currently

Screenshot 2026-03-16 at 5 02 35 PM

Actual behaviour:

 - no options are visible

Screenshot 2026-03-16 at 5 02 51 PM

GowsiSM:
@shiftkey  Thanks for catching that!
The issue was due to `overflow: hidden` on filters-panel, which was clipping the chosen.js dropdown.

I’ve removed that and kept the width fixes on the select. This fixes both the mobile overflow issue and the dropdown visibility problem.

Let me know if anything else needs tweaking!

shiftkey (approved):

PR #5642, opened 2026-03-16
Add pagination to project list

### LINKED ISSUE #4918: Project List missing pagination
_Current Behaviour_ : 
Currently, project list is not paginated. This makes it difficult to navigate and adds alot of scroll time. 

_Image_ : 
ScreenRecording2024-11-18at5 21 01PM-ezgif com-video-to-gif-converter

_Expected Behaviour_ :
A paginated list with a limit of 15 cards (Projects) per page.

I can take it up and make a PR with the necessary changes!

GowsiSM:
Fixes #4918

Currently all projects are rendered on a single page, which leads to excessive scrolling when browsing the list. This change adds client side pagination so the projects are displayed in smaller pages with a limit of 15 projects per page.

Changes made:

* Updated `javascripts/main.js` to slice the filtered project results based on the current page and render pagination controls such as Previous, Next, and numbered page buttons
* Reset the page to 1 whenever any filter changes so filtered results start from the first page
* Added a `#pagination-controls` container in `_includes/scripts.html` which is used to render the pagination buttons

The pagination buttons reuse the existing `radio-btn` and `radio-btn-selected` styles, so no new CSS classes were introduced.


GowsiSM:
Hi @shiftkey,
Fixed the syntax error that was causing the CI failure. The pagination feature should now work as expected, let me know if you'd like any refinements!

shiftkey:
@GowsiSM the pagination buttons could be visually improved a bit:

 - more spacing between the buttons
 - center-aligned

Screenshot 2026-03-21 at 2 53 14 PM

Clicking the buttons also causes the page to scroll upwards, which feels disorienting.

GowsiSM:
@shiftkey, I’ve improved the pagination UI by adding better spacing. I also adjusted the scroll behavior so it now moves smoothly to the project grid instead of jumping to the top. Additionally, I added a small “go to page” input for easier navigation.

All checks are passing now, let me know if anything else should be refined!

shiftkey:
@GowsiSM thanks for this! I'm going to take this in to unblock further iterations...

PR #5693, opened 2026-03-30
Add first contribution

sourav-tz:


shiftkey:
React is already a listed project

https://github.com/up-for-grabs/up-for-grabs.net/blob/gh-pages/_data/projects/react.yml

PR #5763, opened 2026-04-26
fix: include project tags in search filter

### LINKED ISSUE #5514: Search for projects based on tag

GauravPatil2515:
## Summary

Searching for a tag name (e.g. "python", "javascript") returned no results — because `filterProject()` in `search.ts` only matched against `project.name` and `project.desc`. Tags are a primary discovery mechanism on this site, so excluding them from search made the search feel broken.

---

## Problem

```ts
// search.ts — filterProject() before this fix
// only checked name and description, never tags

if (project.name.toLowerCase().indexOf(normalizedSearchText) > -1) {
  return true;
}
if (project.desc.toLowerCase().indexOf(normalizedSearchText) > -1) {
  return true;
}
// tags never checked → user searches "python" → 0 results
```

---

## Fix

Added tag matching as a third check inside `filterProject()`:

```ts
if (
  project.tags?.some(
    (tag) => tag.toLowerCase().indexOf(normalizedSearchText) > -1
  )
) {
  return true;
}
```

**Why `?.some()`:**
- `project.tags` can be `undefined` (not all projects have tags) — optional chaining prevents a runtime crash
- `.some()` short-circuits on first match — no unnecessary iteration

---

## Files Changed

| File | Change |
|------|--------|
| `src/components/data/search.ts` | Added tag matching to `filterProject()` |

---

## Testing

- [x] Search "python" → returns projects tagged with `python`
- [x] Search "javascript" → returns projects tagged with `javascript`  
- [x] Search by project name still works as before
- [x] Search by description still works as before
- [x] Projects with no `tags` field do not crash (optional chaining handles `undefined`)
- [x] Existing tests pass (`npm run test`)

Fixes #5514

### DIFF
--- a/src/components/data/search.ts
+++ b/src/components/data/search.ts
@@ -24,6 +24,10 @@ export function filterProject(
     return true;
   }
 
+  if (project.tags?.some(tag => tag.toLowerCase().indexOf(normalizedSearchText) > -1)) {
+    return true;
+  }
+
   return false;
 }
 
--- a/src/components/search/ProjectEntry.vue
+++ b/src/components/search/ProjectEntry.vue
@@ -7,6 +7,8 @@ const props = defineProps<{
   project: WebsiteProject;
 }>();
 
+const emit = defineEmits<{ 'tag-selected': [tag: string] }>();
+
 const { project } = props;
 const stats = project.stats;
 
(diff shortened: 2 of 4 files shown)


shiftkey:
I can't accept these changes without a good test suite so we can catch future regressions.

PR #5764, opened 2026-04-26
fix: implement AND logic for multi-term search and clickable tag filters

### LINKED ISSUE #250: Filter should be an AND not an OR

### LINKED ISSUE #5514: Search for projects based on tag

### LINKED ISSUE #5759: Feat: Should be able to click on popular tags and apply filter
The list elements provide a very good way to see popular tags. It would improve the user experience if visitors were able to click on them like buttons and automatically add the selected tag to the search filters.

GauravPatil2515:
## Summary

Three related search issues fixed in one focused PR. All changes are in the search layer — no unrelated files touched.

---

### 1. Multi-term search used OR logic, not AND — Issue #250

Searching `"python javascript"` treated the entire string as a literal match, returning 0 results. Users expect multiple words to **narrow** results (AND), not produce no results.

**Before:**
```ts
// Entire string matched as-is — "python javascript" never matches anything
if (project.name.toLowerCase().indexOf(normalizedSearchText) > -1) {
  return true;
}
```

**After:**
```ts
// Split into terms, ALL must match across name, desc, or tags
const terms = searchText.toLowerCase().split(' ').filter((t) => t.length > 0);
if (terms.length === 0) return true;

return terms.every((term) => {
  const inName = project.name.toLowerCase().includes(term);
  const inDesc = project.desc.toLowerCase().includes(term);
  const inTags = project.tags?.some((tag) => tag.toLowerCase().includes(term)) ?? false;
  return inName || inDesc || inTags;
});
```

**Edge cases handled:**
- Empty string → returns all projects (unchanged behaviour)
- Whitespace-only input → `.filter(t => t.length > 0)` strips empty tokens
- Single term → behaves identically to before

---

### 2. Tags excluded from search — Issue #5514

Tag names were never checked in `filterProject()`. Searching `"python"` would miss projects where `python` only appeared in tags.

Fixed by including `project.tags` in the `terms.every()` check above.

---

### 3. Tags not clickable — Issue #5759

Tags rendered as plain `<li>` text with no interaction. Clicking them did nothing.

**Fix:** Wrapped tag content in a `<button>` that emits `tag-selected` → caught in `SearchResults` → sets `searchText` → triggers existing `watch(searchText)` refetch automatically.

```html
<!-- Before -->
<li v-for="tag in project.tags" v-bind:key="tag">{{ tag }}</li>

<!-- After -->
<li v-for="tag in project.tags" v-bind:key="tag">
  <button class="tag-filter" @click="emit('tag-selected', tag)"
    :aria-label="`Filter by tag: ${tag}`">
    {{ tag }}
  </button>
</li>
```

No new state. No new watchers. Reuses existing Vue reactive system entirely.

---

## Files Changed

| File | Change |
|------|--------|
| `src/components/data/search.ts` | AND logic + tag matching in `filterProject()` |
| `src/components/search/ProjectEntry.vue` | Tags → clickable buttons with emit |
| `src/components/search/SearchResults.vue` | Handles `tag-selected` to update `searchText` |

---

## Tests Added (`search.test.ts`)

| Test | Covers |
|------|--------|
| Multi-term AND match | `"python javascript"` matches project with both |
| Multi-term AND fail | `"python ruby"` fails if project only has python |
| Extra whitespace handling | `"  title  "` and `"  title   desc  "` work correctly |
| Tag search single term | `"python"` matches via tags |
| Tag search multi-term | `"python javascript"` matches via tags |

---

## Testing

- [x] `npm run test` — all 67 tests pass
- [x] Searching single term works as before
- [x] Searching `"python javascript"` returns projects matching both
- [x] Searching `"python ruby"` correctly returns 0 results
- [x] Clicking a tag fills search box and filters results
- [x] Tags keyboard navigable (Tab + Enter)
- [x] Visual appearance of tags unchanged

Fixes #250
Fixes #5514
Fixes #5759

### DIFF
--- a/src/components/data/search.ts
+++ b/src/components/data/search.ts
@@ -7,8 +7,6 @@ export function filterProject(
   searchText: string,
   lastUpdated: Date
 ): boolean {
-  const normalizedSearchText = searchText.toLowerCase();
-
   if (
     project.stats.lastUpdated &&
     isBefore(project.stats.lastUpdated, lastUpdated)
@@ -16,19 +14,24 @@ export function filterProject(
     return false;
   }
 
-  if (project.name.toLowerCase().indexOf(normalizedSearchText) > -1) {
+  const terms = searchText
+    .toLowerCase()
+    .split(' ')
+    .filter((t) => t.length > 0);
+
+  if (terms.length === 0) {
     return true;
   }
 
-  if (project.desc.toLowerCase().indexOf(normalizedSearchText) > -1) {
-    return true;
-  }
+  return terms.every((term) => {
+    const inName = project.name.toLowerCase().includes(term);
+    const inDesc = project.desc.toLowerCase().includes(term);
+    const inTags = project.tags?.some((tag) =>
+      tag.toLowerCase().includes(term)
+    ) ?? false;
 
-  if (project.tags?.some(tag => tag.toLowerCase().indexOf(normalizedSearchText) > -1)) {
-    return true;
-  }
-
-  return false;
+    return inName || inDesc || inTags;
+  });
 }
 
 let allProjects: WebsiteProject[] | null = null;
--- a/src/components/data/search.ts
+++ b/src/components/data/search.ts
@@ -24,6 +24,10 @@ export function filterProject(
     return true;
   }
 
+  if (project.tags?.some(tag => tag.toLowerCase().indexOf(normalizedSearchText) > -1)) {
+    return true;
+  }
+
   return false;
 }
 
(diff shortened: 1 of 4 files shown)


shiftkey:
Despite the PR description, I don't seem to see any new tests being added. I can't accept these changes without a good test suite so we can catch future regressions.

PR #5767, opened 2026-04-29
Create skillware.yml

rosspeili:
I think Skillware is a perfect candidate or up for grabs, as it is new, active, clean documentation and contribution guide for new devs and AI agents, and bunch of issues that are easy to solve in less than 30 minutes.


rosspeili:
created fixed version with better tag at #5768

PR #5801, opened 2026-05-14
Add strict validation for project tags and schema compliance

### LINKED ISSUE #2399: Update Yo tool to match current template
Hi Team,

I have recently added my Project [Owasp VulnerableApp](https://github.com/SasanLabs/VulnerableApp/) to the up-for-grabs using YO generator and i found that there are gaps between the tool and the actual requirement of the PR(Build Actions) and hence user has to-fro multiple times as an example:
Tools doesn't have validations or at a minimal text help which mentioned validations like : is not allowed or space in tags is not allowed, lower-case letters are only allowed etc.

I think validations should be added to YO generator tool and if not possible please add text mentioning the same at each input step. I think it is important. 

thanks,
Karan

Reshi12:
## What this fixes
This PR introduces automated validation for project data files to ensure tags follow the required format (lowercase, no spaces, hyphen-separated). It also updates the schema definition to enforce mandatory fields.

## Why it was happening
Contributors were frequently hitting build failures because the `Yo` generator tool lacked input validation, leading to invalid YAML entries in `_data/projects/`. Users were unaware of the specific formatting requirements until the CI pipeline failed, causing unnecessary back-and-forth on PRs.

## What changed
- `scripts/validate_data_files.rb`: Added a `DataValidator` class that performs schema validation via `json_schemer` and enforces regex-based business rules for tags.
- `scripts/schema.json`: Updated the JSON schema to explicitly require `name`, `description`, `tags`, and `upForGrabs` fields, ensuring data consistency across the site.

## How this was tested
- Ran the validation script locally against existing project files to ensure no regressions.
- Verified that invalid tags (e.g., "Web Security") trigger a descriptive error message and exit with a non-zero status code.
- Confirmed the CI pipeline correctly identifies and blocks malformed entries.

## Impact
Without this, the repository will continue to accumulate malformed project data, leading to inconsistent UI rendering and broken build actions. This change shifts the validation feedback loop left, catching errors before they reach the PR stage.

Closes #2399

### DIFF
--- a/scripts/validate_data_files.rb
+++ b/scripts/validate_data_files.rb
@@ -1,11 +1,56 @@
 # frozen_string_literal: true
 
-require 'up_for_grabs_tooling'
+require 'yaml'
+require 'json'
+require 'json_schemer'
+require 'pathname'
 
-root = Pathname.new(File.dirname(__dir__))
+# Validates project data files against schema and business rules.
+# Business rules:
+# 1. Tags must be lowercase and contain no spaces.
+# 2. Project files must be valid YAML.
+class DataValidator
+  def initialize
+    @schema = JSON.parse(File.read('scripts/schema.json'))
+    @schemer = JSONSchemer.schema(@schema)
+  end
 
-result = CommandLineValidator.validate(root)
+  def validate_file(file_path)
+    data = YAML.safe_load(File.read(file_path))
+    
+    # Schema validation
+    unless @schemer.valid?(data)
+      errors = @schemer.validate(data).map { |e| e['data_pointer'] }.join(', ')
+      puts "Validation failed for #{file_path} at: #{errors}"
+      return false
+    end
 
-CommandLineFormatter.output(result)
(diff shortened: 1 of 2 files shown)


scripts/validate_data_files.rb
```diff
@@ -1,11 +1,56 @@
...
+require 'json'
+require 'json_schemer'
+require 'pathname'
 
-root = Pathname.new(File.dirname(__dir__))
+# Validates project data files against schema and business rules.
+# Business rules:
+# 1. Tags must be lowercase and contain no spaces.
+# 2. Project files must be valid YAML.
+class DataValidator
```
shiftkey, line 12: - https://github.com/up-for-grabs/tooling/blob/main/lib/validators/command_line.rb
 - https://github.com/up-for-grabs/tooling/blob/main/schema.json

There's already a validator that does schema changes - please look that over and contribute any additional changes you think are needed there (instead of clobbering it with this).

PR #6101, opened 2026-08-31
Update Bernstein project description

chernistry:
Bernstein's project description changed, and the `desc` field in `_data/projects/bernstein.yml` still carried the previous wording. This updates it to match the current description; no other fields are touched.


chernistry:
Checks are green and the description now matches what's shown in-repo. Happy to update again if anything else drifts — let me know if there's anything else needed on this end.

